PR #56, opened 2019-11-16
Offline registry reader

AlekStoyanov:
Offline hive parsing functionality, with Zimmerman's library.


WPF/SeeShells/SeeShells/ShellParser/OfflineRegistryReader.cs
```diff
@@ -1,15 +1,138 @@
...
+
         public List<RegistryKeyWrapper> GetRegistryKeys()
         {
-            throw new NotImplementedException();
+            List<RegistryKeyWrapper> retList = new List<RegistryKeyWrapper>();
+
+            foreach (string location in Parser.GetLocations())
+            {
+                // Location of Offline Hive (to be obtained through UI)
+                var hive = new RegistryHiveOnDemand(@"C:\Users\Aleksandar\Desktop\CurrUser");
```
klaki892: We cant keep this, this'll have to become a variable thats probably passed into the registry reader. It'll break immediately on anyone else pulling the solution

WPF/SeeShells/SeeShells/ShellParser/OfflineRegistryReader.cs
```diff
@@ -1,15 +1,138 @@
...
+
+                foreach (byte[] keyValue in IterateRegistry(hive.GetKey(@"ROOT\" + location), hive, location, 0, ""))
```
klaki892: @AlekStoyanov  does `"Root\"` work on both exported registries and offline hives?

WPF/SeeShells/SeeShellsTests/ShellParser/OfflineRegistryReaderTests.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace SeeShellsTests.ShellParser
+{
+    [TestClass]
+    public class OfflineRegistryReaderTests
+    {
+        /// <summary>
+        /// Tests that the Offline Registry Reader can read from a hive file.
+        /// </summary>
+        [TestMethod]
+        public void GetRegistryKeysTest()
```
klaki892: As of right now this test will fail on anyone else's machine, we probably need a test resource folder with a default NTUser file so that this test can run without external dependencies.

WPF/SeeShells/SeeShells/ShellParser/OfflineRegistryReader.cs
AlekStoyanov: No, as we have tested. We'd have to check the file input and get the correct path from the config parser. For now I'll move the ROOT part in the OfflineMockConfigParser.

WPF/SeeShells/SeeShells/ShellParser/OfflineRegistryReader.cs
```diff
@@ -1,15 +1,139 @@
...
+
         public List<RegistryKeyWrapper> GetRegistryKeys()
         {
-            throw new NotImplementedException();
+            List<RegistryKeyWrapper> retList = new List<RegistryKeyWrapper>();
+
+            foreach (string location in Parser.GetLocations())
+            {
+                // Location of Offline Hive
+                var hive = new RegistryHiveOnDemand(Directory.GetParent(Directory.GetCurrentDirectory()).Parent.FullName + @"\TestResource\NTUSER.DAT");
```
klaki892: We can't have a hard reference like `@"\TestResource\NTUSER.DAT"` in the normal code thats meant for usage by the main program. 
the path to the file should be determined by something else, not OfflineRegistReader. Some other part of the code should be passing in the route for the file.
This will probably require an additional parameter in the OfflineRegistryReader's constructor that something similar to Program.cs has to pass into it.
```suggestion
                var hive = new RegistryHiveOnDemand(registryFile);
```

WPF/SeeShells/SeeShells/ShellParser/OfflineRegistryReader.cs
```diff
@@ -1,15 +1,139 @@
...
-    class OfflineRegistryReader : IRegistryReader
+    /// <summary>
+    /// This class is used to read an offline regisry hive file.
+    /// Using the Registry API, from EricZimmerman, it reads registry values from an hive file. 
+    /// </summary>
+    public class OfflineRegistryReader : IRegistryReader
     {
+        IConfigParser Parser { get; }
+
+        public OfflineRegistryReader(IConfigParser parser)
```
klaki892: As mentioned below, something should be passed into OfflineRegistryReader so that it doesnt have to obtain the file by itself, that way its not the Reader's job to handle file related management other than reading the file.
```suggestion
        public OfflineRegistryReader(IConfigParser parser, File registryFile)
```
```suggestion
        public OfflineRegistryReader(IConfigParser parser, String registryFile)
```

Issue #60, opened 2020-01-08
API - Setting up the Environment


klaki892:
Using Heroku, Setup the Node.js /Express.Js server to accept connections and be able to preform interaction with the database

Acceptance Criteria:

- A testing endpoint can be hit from a web browser 
- Communication with the Heroku Postgres database can be established for future endpoint usage.

PR #65, opened 2020-01-10
Add basic files and code for API

szczura:
This commit adds the node modules, package.json, and some basic index.js code for a start to see if it deploys to heroku correctly. #60


szczura:
I set up the buildpack Klayton suggested and set the PROJECT_PATH=api

Issue #51, opened 2019-11-04
WPF interpreting and classifying ShellBag items


szczura:
Given ShellBag binary data, identify ShellItem and corresponding parameters to known types and prepare it for consumption for various output methods.
If an an unknown ShellItem is encountered it should be logged to an error log.
- Target Fields
  - File Path
  - Date (created, accessed, modified)

Acceptance Criteria:
- Converts ShellBag binary data to an object with target fields.
- If an an unknown ShellItem is encountered it should be logged to an error log.

PR #67, opened 2020-01-11
Add, Refactor, and Expand type indicator-base ShellItems

klaki892:
Created Skeleton of new ShellItemList for ShellItem distribution
Refactors ShellItem classes and Expansion Block implementations
Expanded ShellItem Definitions to obtain more properties
Adds tests for known ShellItem examples
full enumeration of ExpansionBlock properties still pending
Adds IntelliJ user config to .gitignore

Provides Significant progress to #51


WPF/SeeShells/SeeShells/ShellParser/ShellItems/ShellItemList.cs
```diff
@@ -0,0 +1,51 @@
...
+        public ShellItemList(byte[] buf, int offset) : base(buf, offset) { }
+
+        protected  IShellItem GetItem(int off)
+        {
+            //item to return
+            IShellItem item = null;
+            //the shell item type which can be identified by 2 bytes (0xXX)
+            int type = unpack_byte(off + 2);
+
+            //TODO: Aleks - Dynamically call upon correct ShellItem0xXX to be instantiated
```
klaki892: Pointing this out for @AlekStoyanov 

We messed up in our original design. It turns out that the `IRegistryReader.getValue() Will not return a ShellItem ready to be decoded but instead a [Shell Item List].(https://github.com/libyal/libfwsi/blob/master/documentation/Windows%20Shell%20Item%20format.asciidoc#2-shell-item-list)

Basically you'll need to create a new `ShellItemList` for each registry value, and implement this method so you can ultimately return the list of `IShellItem`s to the greater good.

WPF/SeeShells/SeeShells/ShellParser/ShellItems/ShellItemList.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            return item;
+        }
```
klaki892: Pointing this out for @AlekStoyanov 

We messed up in our original design. It turns out that the `IRegistryReader.getValue()` Will not return a ShellBag Item ready to be decoded but instead a [Shell Item List](https://github.com/libyal/libfwsi/blob/master/documentation/Windows%20Shell%20Item%20format.asciidoc#2-shell-item-list) object.

Basically you'll need to create a new `ShellItemList` for each registry value, and implement this method so you can ultimately return the list of `IShellItem`s to the greater good.

WPF/SeeShells/SeeShells/ShellParser/ShellItems/Block.cs
```diff
@@ -2,21 +2,21 @@
...
-        public Block(byte[] buf, int offset, object parent)
+        protected byte[] buf { get; set; }
+        public int offset { get; protected set; }
+        protected object parent { get; set; }
+        protected Block(byte[] buf, int offset)
         {
             this.buf = buf;
             this.offset = offset;
-            this.parent = parent;
+            parent = parent;
```
szczura: isn't parent not changing here?

klaki892: Yup looks like a useless self-assignment to me. 
Happened while taking parent out of the constructor.
fixed.

WPF/SeeShells/SeeShells/ShellParser/ShellItems/ShellItemList.cs
AlekStoyanov: By `IRegistryReader.getValue()` do you mean `IRegistryReader.GetRegistryKeys()` ?

So, my understanding is, the `ShellBagParser` `GetShellItems` method should create an instance of `ShellItemList` passing it the raw byte array from the `IRegistryReader` and call `Items` method in order to obtain a list of `IShellItem`. Meanwhile all the Activator reflection magic should happen in the `GetItem` method in `ShellItemList`

klaki892: Yes, everything you've said is correct.

Issue #61, opened 2020-01-08
API - Register API Endpoint


klaki892:
As a developer, This endpoint is needed to register administrators on the SeeShells website.
Register an administrator requires secure salting and hashing of the password 

Acceptance Criteria:
- Upon hitting the Register Endpoint, a Username and Salted Password are created in the database
- an error is propagated to the user if an error occurs during the registration of the user

PR #69, opened 2020-01-13
Register endpoint

szczura:
#68 This endpoint seems to be working properly; users are created when I use postman. I'll see if the hash & salt parts are working properly when I'm working on the login endpoint.


api/package.json
```diff
@@ -3,9 +3,7 @@
   "version": "1.0.0",
   "description": "",
   "main": "index.js",
-  "directories": {
-    "doc": "docs"
-  },
+  "directories": {},
   "scripts": {
```
klaki892: How are you testing this locally? Right now the `pg`  library isn't defined as a requirement in the `package.json`. 
This should cause missing dependency issues when someone preforms an `npm install`

klaki892:
Fixes #61

api/package.json
szczura: Good catch, forgot to commit the updated package file

PR #75, opened 2020-01-15
Create WPF GUI

bridCquinn:
Home
Timeline Page (Not the actual timeline)
Help


klaki892:
Since this PR hasnt been merged,
If not having the GUI does not block another issue, maybe #92 can also be incorporated before merging since it affects the exact same files (and is a bugfix/improvement on these proposed files)

szczura:
I think Klayton's idea works well; we can revisit this PR on Wednesday or whenever the GUI gets updated before then.

szczura:
Note: make the parse button on the home page nice and big

WPF/SeeShells/SeeShells/IO/HtmlIO.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Web.UI;
+
+namespace SeeShells.IO
+{
+    /// <summary>
+    /// This class is used to manage the output of ShellBags data in the form of a printable HTML timeline.
+    /// </summary>
+    public static class HtmlIO
```
klaki892: Html IO is a whole different issue and shouldn't be worked on at the same time as this UI issue. 

I'd really prefer it if this was a separate pull request.
If this is 100% done (and based on the contents of the html writer it doesn't appear so) and requires no further changes, then change the title of the PR and the description to accommodate that, else make a new PR for anything html IO specific, and pull the respective code out of this PR so it can be merged and completed as it's own issue

Issue #83, opened 2020-01-15
Update IShellItem documentation


szczura:
Add descriptions to what properties are returned in the interface.

PR #70, opened 2020-01-13
Handle empty date parsing in ShellBag parsing

klaki892:
Fixes issue where an exception would be thrown on invalid DateTime conversion

Credit to @AlekStoyanov for seeing the original `Block.cs` caused an exception when only 0x00 bytes got parsed.

PR #94, opened 2020-01-16
WPF interpreting and classifying ShellBag items #51

AlekStoyanov:


WPF/SeeShells/SeeShells/ShellParser/ShellBagParser.cs
```diff
@@ -20,8 +21,19 @@ public ShellBagParser(IRegistryReader registryReader)
         /// <returns>a list of different variety ShellBag items</returns>
         public List<IShellItem> GetShellItems()
         {
-            // TODO: actual logic
-            return new List<IShellItem>();
+            List<IShellItem> shellItems = new List<IShellItem>();
+            foreach (RegistryKeyWrapper keyWrapper in registryReader.GetRegistryKeys())
+            {
+                if(keyWrapper.getValue() != null && keyWrapper.getValue().Length > 6) // Some Registry Keys are null
```
AlekStoyanov: @klaki892 I need your help with figuring out what exactly the issue is with the length of one of our raw binary arrays from the offline hive.
Take out the `&& keyWrapper.getValue().Length > 6` from the if statement and run the `GetShellItemsOffineTest` test.
My understanding is that the offset causes the unpacking method to start reading out of bounds but I can't quite put my finger on what exactly is going on.

klaki892: You're initial idea is correct, the offset is causing the upacking to read out of bounds
After debugging, the issue occurs when a byte array of `0x03 0x00 0x00 0x00` is passed to the ShellItemList. 
Its quite simple, as the code is written line 59 of `ShellItemList.cs` will add the size of the last item to the offset. We Simply dont check if the offset is equal to the size of the byte array before looking up the next size of the item. 
The fix is simply to put the following  lines ahead of line 47 in `ShellItemList`:
```
               //prevent out of bounds reading when the exact size of an item meets the next expected offset
                if (size == off && off != 0)
                    break;
```
The offline test case passed with this change and the one mentioned below.

Also dont do length < 6 there are actually valid shellitems that might only have 4 bytes (see ShellItem0x00) ShellItems themselves will tell you if they are empty (by size being 00)
```suggestion
                if(keyWrapper.getValue() != null ) // Some Registry Keys are null
```

WPF/SeeShells/SeeShells/SeeShells.csproj
```diff
@@ -40,17 +40,23 @@
...
       <HintPath>resources\CommandLine.dll</HintPath>
     </Reference>
+    <Reference Include="Microsoft.CSharp" />
     <Reference Include="NFluent">
       <HintPath>resources\NFluent.dll</HintPath>
     </Reference>
-    <Reference Include="NLog">
-      <HintPath>resources\NLog.dll</HintPath>
+    <Reference Include="NLog, Version=4.0.0.0, Culture=neutral, PublicKeyToken=5120e14c03d0593c, processorArchitecture=MSIL">
+      <HintPath>..\packages\NLog.4.6.8\lib\net45\NLog.dll</HintPath>
```
klaki892: If you're going to get the package from NuGet (which is what appears to have happened here, remove the NLog.Dll that we manually include in our resources folder.

WPF/SeeShells/SeeShells/ShellParser/ShellItems/Block.cs
```diff
@@ -81,7 +82,15 @@ protected string unpack_wstring(int off, int length = 0)
...
                 }
                 while (buf[offset + off + length - 2] == 0 && buf[offset + off + length - 1] == 0) length -= 2;
-                return Encoding.Unicode.GetString(buf, offset + off, length);
+                try
+                {
+                    return Encoding.Unicode.GetString(buf, offset + off, length);
+                }
+                catch(Exception ex)
+                {
+                    logger.Error(ex, "Could not resolve dir name\n" + ex.ToString());
```
klaki892: Misleading error message. unpack_wstring is used for more than just figuring out directory names, its meant for all strings we come across (e.g. a filename representation in the Extension block BEEF0004)

WPF/SeeShells/SeeShells/ShellParser/ShellItems/Block.cs
```diff
@@ -164,7 +173,15 @@ protected DateTime unpack_dosdate(int off)
...
-                return new DateTime(year, month, day, hour, minute, sec);
+                try
+                {
+                    return new DateTime(year, month, day, hour, minute, sec);
+                }
+                catch(Exception ex)
+                {
+                    logger.Error(ex, "Could not resolve DateTime\n" + ex.ToString());
+                }
+                return DateTime.MinValue; ;
```
klaki892: 1. We usually dont want to do a catch-all for `Exception` ... ever. [DateTime's documentation states ](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.-ctor?view=netframework-4.8#System_DateTime__ctor_System_Int32_System_Int32_System_Int32_System_Int32_System_Int32_System_Int32_)it will only return an `ArgumentOutOfRange` Exception.

2. If you are going to do this  exception handling for anything in the constructor, theres no reason for our above error checking `                if ((dosdate == 0  || dosdate == 1) && dostime == 0)
`

WPF/SeeShells/SeeShells/ShellParser/ShellItems/ShellItemList.cs
```diff
@@ -7,22 +7,35 @@ namespace SeeShells.ShellParser.ShellItems
...
+            {
+                // Getting here means that the ShellItem is unidentified
+                logger.Info("Could not identify ShellItem 0x" + postfix);
+                return new ShellItem(buf);
+
+                // TODO Embedded Scripting
+            }
 
-            return item;
+            return (IShellItem)Activator.CreateInstance(type, buf);
```
klaki892: We need error logging here if the construction of a ShellItem fails. And probably should output the byte array that caused it.

szczura:
Klayton has been very thorough here; I don't have anything to add until those changes are made.

WPF/SeeShells/SeeShells/ShellParser/ShellBagParser.cs
AlekStoyanov: Fixed!

WPF/SeeShells/SeeShells/SeeShells.csproj
AlekStoyanov: Fixed!

WPF/SeeShells/SeeShells/ShellParser/ShellItems/Block.cs
AlekStoyanov: We are not doing exception logging at that level anymore.

WPF/SeeShells/SeeShells/ShellParser/ShellItems/Block.cs
AlekStoyanov: We are not doing exception logging at that level anymore.

WPF/SeeShells/SeeShells/ShellParser/ShellItems/ShellItemList.cs
AlekStoyanov: Added, however I am catching a general `Exception` even though it's not the best practice, since there is a multitude of exceptions that can be thrown at this stage and also as we start to fix errors in parsing we might add more exceptions to be thrown which we'll have to keep track of if it's not a generic `Exception`..
I am logging what the exception is, so it will be known when the time for debugging comes.

WPF/SeeShells/SeeShells/ShellParser/ShellItems/Block.cs
```diff
@@ -159,12 +150,6 @@ protected DateTime unpack_dosdate(int off)
                 ushort dosdate = (ushort)(buf[offset + off + 1] << 8 | buf[offset + off]);
                 ushort dostime = (ushort)(buf[offset + off + 3] << 8 | buf[offset + off + 2]);
 
-                //check if the bytes contained no data
-                if ((dosdate == 0  || dosdate == 1) && dostime == 0)
-                {
-                    return DateTime.MinValue; //same thing as invalid. (minvalue goes below the epoch)
-                }
-
```
klaki892: This shouldn't have been removed.
This is an actual part of parsing as detailed in #70 
For this shell bag to be parsed, the  value of DateTime.MinValue is already documented as a possibility in #83 
Removing this introduces errors into parsed shellbags. Without this parsing section we permanently remove the ability to read the ShellItem as detailed in the test case which accompanied it. 

I know we've discussed the contention with providing base values for parsing conditions that throw exceptions, but this is fixing the limitation of the DateTime class itself, where it is unable to represent a time of 0, and instead replaces it with the closest acceptable value of 0.

WPF/SeeShells/SeeShellsTests/ShellParser/ShellItems/ShellItemsTests.cs
```diff
@@ -183,31 +183,5 @@ public void ShellItem0x71FTPTest()
...
-            ShellItem0x31 shell = new ShellItem0x31(value);
-            printProperties(shell);
-
-            //Test passes by throwing no Exceptions.
-        }
-
-
-
-
-
```
klaki892: Explanation of why this shouldn't have been removed is in other related comment

WPF/SeeShells/SeeShells/ShellParser/ShellItems/Block.cs
AlekStoyanov: I understand

PR #98, opened 2020-01-20
A skeleton of CSV IO (HTML IO should be similar)

AlekStoyanov:
This is my idea of static classes to be used for CSV and HTML IO

### DIFF
--- a/WPF/SeeShells/SeeShells/IO/CsvIO.cs
+++ b/WPF/SeeShells/SeeShells/IO/CsvIO.cs
@@ -23,9 +23,9 @@ namespace SeeShells.IO
         /// <summary>
         /// Creates a ShellItem list from ShellBag data parsed from a CSV.
         /// </summary>
-        /// <param name="dir">The directory of a source CSV file</param>
+        /// <param name="filePath">The path to a source CSV file</param>
         /// <returns>A list of ShellItems</returns>
-        public static List<IShellItem> ImportCSVFile(String dir)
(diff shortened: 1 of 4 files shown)


WPF/SeeShells/SeeShells/IO/CsvIO.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <returns></returns>
+        public static void OutputCSVFile(List<IShellItem> ShellItems, String dir)
+        {
+
+        }
+
+        /// <summary>
+        /// Creates a ShellItem list from ShellBag data parsed from a CSV.
+        /// </summary>
+        /// <param name="dir">The directory of a source CSV file</param>
```
klaki892: Are you sure you only want the directory and not a specific file to be indicated?

Only taking the directory implies you will parse all csv's In a given directory. Which is fine if that's what you're going for.

AlekStoyanov: The filename should be attached at the end of the directory. Like`C:\Users\User\Desktop\TheCsvFile.csv`

klaki892: You should probably change that variable name or description to better represent that. we all know you're coding this, but to the person who implements this on the gui side might pass you an entire directory instead of a file. 'filepath' instead of 'dir'  would make that clearer

AlekStoyanov: You're right.

Issue #116, opened 2020-01-25
Add OS (Different registry file locations than previous) API Endpoint


szczura:
When new versions of Windows comes out, this endpoint can be used to add them to the database; this is needed for the Get OS & Registry Locations API Endpoint to stay relevant.

This is a POST endpoint that takes the OS version number, the OS name, and an array of file locations for the registry keys.

Issue #103, opened 2020-01-22
Add OS (same keys as previous) API Endpoint


szczura:
When new versions of Windows comes out, this endpoint can be used to add them to the database; this is needed for the Get OS & Registry Locations API Endpoint to stay relevant.

This is a POST endpoint that takes in the OS version number, the OS name, and what key grouping the OS uses.

Issue #104, opened 2020-01-22
Get Possible Registry Locations API Endpoint


szczura:
The website will need this endpoint to list the possible registry locations when allowing an admin to add a new OS. This GET endpoint will return an array of the groupings of registry locations in the database and the location ID that the OS versions point to.

PR #118, opened 2020-01-28
Add OS API Endpoints

szczura:
These 2 endpoints let us add new OS versions. The regular add OS endpoint lets us add a new version that uses the same registry key locations as a previous version (#103 ). The add OS with file locations endpoint lets us add a new version that has different registry key locations (#116 ).


szczura:
Just committed an endpoint for issue #104

PR #122, opened 2020-02-03
Node, NodeCollection and NodeParser Skeleton

AlekStoyanov:
Introduces the following classes:
Node -> Formerly called Pretty Object, containing graphical objects to be displayed on a timeline.
NodeCollection -> Formerly called PrettyObjectCollection, used to store a list of nodes and be globally accessible from the App class
NodeParser -> Used to populate the the list of nodes in the NodeCollection from a list of Events

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Node/Node.cs
+++ b/WPF/SeeShells/SeeShells/UI/Node/Node.cs
@@ -3,20 +3,27 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using System.Windows.Shapes;
 
 namespace SeeShells.UI.Node
 {
     public class Node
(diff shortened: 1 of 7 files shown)


WPF/SeeShells/SeeShells/UI/Node/Node.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Threading.Tasks;
+using System.Windows.Shapes;
+
+namespace SeeShells.UI.Node
+{
+    public class Node
+    {
+        IEvent aEvent;
+        Ellipse dot;
+        Rectangle block;
```
klaki892: Will these be properties be publicly accessible?

The filters need someway to alter the visibility property of `dot` and also to get the information from `aEvent`

PR #125, opened 2020-02-05
Event parser

yara-58:


WPF/SeeShells/SeeShells/UI/EventParser.cs
```diff
@@ -0,0 +1,119 @@
...
+        public EventParser(List<IShellItem> shells)
+        {
+            foreach (IShellItem item in shells)
+            {
+                Event eM = createModifiedEvent(item);
+                Event eA = createAccessedEvent(item);
+                Event eC = createCreationEvent(item);
+                eventList.Add(eM);
+                eventList.Add(eA);
+                eventList.Add(eC);
```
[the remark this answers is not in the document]
szczura: Each shellitem also won't have all the different time event types. We need to check what type it is before creating an event for it to add to the list.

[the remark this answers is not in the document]
klaki892: > Each shellitem also won't have all the different time event types. We need to check what type it is before creating an event for it to add to the list.

It's impossible to check the type because there's more types than we know.
But you are right, each shellitem wont necessarily even have the original 3. So a check should be done if the value is the `DateTime.MinValue` (1/1/0001) before making it into an event.

WPF/SeeShells/SeeShells/UI/EventParser.cs
```diff
@@ -1,23 +1,119 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
 namespace SeeShells.UI
 {
-    public class EventParser : Block, IShellItem
+    class EventParser
```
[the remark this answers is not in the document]
AlekStoyanov: I think the only solution is to make it public, since this will be called on from somewhere to populate the `EventCollection` class. Meaning that unit testing is not our only concern.

WPF/SeeShells/SeeShellsTests/UI/EventParserTest.cs
```diff
@@ -0,0 +1,45 @@
...
+            List<IShellItem> shellItems = new List<IShellItem>();
+            Dictionary<string, string> shellItemProperties = new Dictionary<string, string>();
+            shellItemProperties.Add("Size", "0");
+            shellItemProperties.Add("Type", "31");
+            shellItemProperties.Add("TypeName", "Some Type Name");
+            shellItemProperties.Add("Name", "Some Name");
+            shellItemProperties.Add("ModifiedDate", "1/1/0001 12:00:00 AM");
+            shellItemProperties.Add("AccessedDate", "1/1/0001 12:00:00 AM");
+            shellItemProperties.Add("CreationDate", "1/1/0001 12:00:00 AM");
+            shellItemProperties.Add("LastAccessedDate", "1/1/001 12:00:00 AM");
```
klaki892: This should fail in a test case. No events should be made from this because 1/1/0001 is not a real file date. (Windows respects the epoch time, 1/1/1970)

WPF/SeeShells/SeeShells/UI/EventParser.cs
```diff
@@ -0,0 +1,50 @@
...
+        public List<IEvent> Parser()
+        {
+            foreach (IShellItem item in shells)
+            {
+                IDictionary<String, String> parser = item.GetAllProperties();
+                foreach(var el in parser)
+                {
+                    String check = el.Key;
+                    String date = "Date";
+                    if (check.Contains(date))
```
klaki892: This isnt the only check that should be performed
If you see a `DateTime.MinValue` (1/1/0001) you shouldnt make an event from this because it's a default value similar to `null` indicating there wasnt a real date.

```suggestion
                    if (check.Contains(date) && el.Value != DateTime.MinValue)
```

yara-58: oh got you makes sense

WPF/SeeShells/SeeShells/UI/EventParser.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// creates the eventList that will be returned full of events based on ShellItems
+        /// </summary>
+        List<IEvent> eventList = new List<IEvent>();
+        List<IShellItem> shells;
+        /// <summary>
+        /// Constructor for the EventParser that loops through each ShellItem and creates an event 
+        /// for each Modified,Accessed and Creation date for that shellItem.
+        /// </summary>
+        /// <param name="shells"></param>
+        public EventParser(List<IShellItem> shells)
```
AlekStoyanov: Why does this still have a constructor ?
Make the class static.

WPF/SeeShells/SeeShells/UI/EventParser.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            this.shells = shells;
+        }
+        public List<IEvent> Parser()
```
AlekStoyanov: This method should be static and take the `List<IShellItem>`.
Also the name of the methods should be, or contain verbs, so change it to `GetEvents`. (this is to maintain consistency with the NodeParser class naming scheme)

WPF/SeeShells/SeeShells/UI/EventParser.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            foreach (IShellItem item in shells)
+            {
+                IDictionary<String, String> parser = item.GetAllProperties();
+                foreach(var el in parser)
+                {
+                    String check = el.Key;
```
AlekStoyanov: I don't think it is necessary to capture these values, it would be easier to read if you use them straight up.
Ex. `el.Key.Contains("Date")`

WPF/SeeShells/SeeShells/UI/EventParser.cs
```diff
@@ -0,0 +1,50 @@
...
+                    String date = "Date";
+                    if (check.Contains(date) && Convert.ToDateTime(el.Value) != DateTime.MinValue)
+                    {
+                        String name = item.Name;
+                        String typename = item.TypeName;
+                        DateTime eventDate = Convert.ToDateTime(el.Value);
+                        String[] type = check.Split('D');
+                        String eventType = type[0];
+                        Event e = new Event(name, eventDate, item, eventType);
```
AlekStoyanov: It seems the Event class only takes `name, eventDate, item, eventType`, then why are you capturing variables that are then never used, such as `typename` and `type` ?
Also as mentioned before, you don't have to capture the values in variables, you can use them straight up.

klaki892: +1 for typename, but the first half of `type` is used to generate eventType.
You could eliminate the variable, but that's just stylistic choice

WPF/SeeShells/SeeShells/UI/EventParser.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Threading.Tasks;
+
+namespace SeeShells.UI
+{
+    public static class EventParser
+    {
+        /// <summary>
+        /// creates the eventList that will be returned full of events based on ShellItems
+        /// </summary>
+        List<IEvent> eventList = new List<IEvent>();
```
szczura: I think you need to make this static.

The build is failing with this message:
"UI\EventParser.cs(15,22): error CS0708: 'EventParser.eventList': cannot declare instance members in a static class"

PR #130, opened 2020-02-07
Home page refactored & Tab Control turned into Pages & Icon Added

### LINKED ISSUE #128: Refactor the tabs into pages
GUI update so that people can't click the timeline tab without having clicked parse.

szczura:
Essentially resolves #128 because I threw all the XAML into pages, but it needs some cleaning up on the pages besides the home page.

Making the PR so that people can work with the pages instead of tab items if needed, but I will continue working on the GUI.

Also added an icon.


WPF/SeeShells/SeeShells/ShellParser/Script.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using NLua;
+
+namespace SeeShells.ShellParser
+{
+    class Script
```
klaki892: Was there supposed to be Script stuff in this PR?

szczura: nope lol I think when I switched branches VS messed up

WPF/SeeShells/SeeShells/UI/Templates/CheckBoxWithLabel.xaml.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Windows.Media.Imaging;
+using System.Windows.Navigation;
+using System.Windows.Shapes;
+
+namespace SeeShells.UI.Templates
+{
+    /// <summary>
+    /// Interaction logic for CheckBoxWithLabel.xaml
+    /// </summary>
+    public partial class CheckBoxWithLabel : UserControl
```
klaki892: Does this still fire off the events from checkboxes? 
e.g. the [onChecked event](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.checkbox?view=netframework-4.8#events)

szczura: we can bind to the isChecked label to fire off the events. it'll essentially be the same thing as using onChecked, just written a little different. Also, the checkboxes currently weren't doing anything.

szczura:
Removed the one script file.
The icon is just a setting in VS; you don't even have to write any code so I figured it wasn't worth mentioning. It takes like 5 seconds to change it.

WPF/SeeShells/SeeShells/UI/Templates/CheckBoxWithLabel.xaml.cs
szczura: If subscribing to isChecked is too difficult, its super easy to change it back anyway.

Issue #129, opened 2020-02-05
Make the parse button do its thing


szczura:
Enable the parse button on the GUI.
Starting the Shellbag parsing, and passing the information to the rest of the GUI.

Acceptance Criteria: Once the user clicks parse all relevant actions are preformed (or started) to allow the timeline to display information.

PR #131, opened 2020-02-09
Added functionality to the browse/update buttons on the home page

### LINKED ISSUE #30: WPF download configuration files from database
As an investigator or regular user, I want to download configuration files from a database to parse ShellBags.
- Hit the OS & Locations endpoint and parse the JSON object
- Hit the GUID endpoint and parse the JSON object
- Hit the Scripts endpoint and parse the JSON

### LINKED ISSUE #31: WPF overwrite local configuration files
As an investigator or regular user, I want to overwrite local configuration files with the database configuration files.
- Take the parsed JSON data from #30 and overwrite the configuration files

szczura:
This resolves #30 (with the exception of the scripting file since the whole scripting thing is still being worked on) and resolves #31 

(Also added a help button on the home page)


WPF/SeeShells/SeeShells/UI/Pages/Home.xaml.cs
```diff
@@ -32,37 +49,169 @@ private void OfflineBrowseButton_Click(object sender, RoutedEventArgs e)
...
+                MessageBox.Show("Select a proper GUID configuration file or create a new one.", "Missing GUID File", MessageBoxButton.OK, MessageBoxImage.Error);
+                return;
+            }
+            if (!File.Exists(locations.ScriptFileLocation))
+            {
+                MessageBox.Show("Select a proper script configuration file or create a new one.", "Missing Script File", MessageBoxButton.OK, MessageBoxImage.Error);
+                return;
+            }
+
+
```
klaki892: This might need to be addressed in #129, but right now this prevents the user from using the program while offline.
(e.g.  user downloads the program, transfers to a flashdrive, then uses the program on an offline computer)
We need to have defaults embedded in the program rather than throwing errors on no files found

PR #141, opened 2020-02-12
Update Parse Button to work with Timeline

klaki892:
- change debug MessageBox to logged message
- refactored error messageboxes to common function
- fixes null check on OSVersion

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Pages/Home.xaml.cs
+++ b/WPF/SeeShells/SeeShells/UI/Pages/Home.xaml.cs
@@ -12,7 +12,6 @@ using SeeShells.UI.ViewModels;
 using Newtonsoft.Json;
 using SeeShells.IO.Networking.JSON;
 using SeeShells.ShellParser.ShellItems;
-using System.Threading;
 using SeeShells.ShellParser;
 using SeeShells.UI.Node;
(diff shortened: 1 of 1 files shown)


WPF/SeeShells/SeeShells/UI/Pages/Home.xaml.cs
```diff
@@ -246,18 +247,14 @@ private async void ParseButton_Click(object sender, RoutedEventArgs e)
             stopwatch.Start();
             App.ShellItems = await ParseShellBags();
             List<IEvent> events = new List<IEvent>(); //FIXME EventParser.getEvents(shellItems);
-            App.nodeCollection.nodeList = new List<Node.Node>(); //TODO REMOVE ME once merged with filtering code, unneccesarly NPE stopping
             App.nodeCollection.nodeList.AddRange(NodeParser.GetNodes(events));
```
AlekStoyanov: Why are you using `AddRange()` instead of just setting it equal to `NodeParser.GetNodes(events)` ?

klaki892: No particular strong reason, nodelist couldve been initalized and populated and this wouldnt destroy the old reference.
It can be switched.

PR #146, opened 2020-02-17
Embedded scripting

### LINKED ISSUE #79: Make Lua script to parse known shell item
To complete:
- make a script that attempts to parse a shell item
- the results it parses are the same as what the shellbag parser gives us

szczura:
Still a work in progress but the Control Panel script gets the properties it needs! Resolves #79 

I want to know what you guys think about what I talk about on line 161 in LuaShellItem.cs
Since the unpack functions that were made are protected in the Block class, they can't be used from the Lua scripts. We can change those protected functions to public, copy over the functions into the LuaShellItem class, or force the scripts to do that work themselves (or some other solution I didn't think of).

### DIFF
--- a/WPF/SeeShells/SeeShells/App.xaml.cs
+++ b/WPF/SeeShells/SeeShells/App.xaml.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Configuration;
 using System.Data;
 using System.Windows;
+using SeeShells.ShellParser.Scripting;
 
 namespace SeeShells
 {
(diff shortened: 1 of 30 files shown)


WPF/SeeShells/SeeShells/App.xaml.cs
```diff
@@ -28,5 +29,10 @@ public partial class App : Application
         /// Collectin of <see cref="ShellItem"/> which is populated after a parsing operation.
         /// </summary>
         public static List<IShellItem> ShellItems { get; set; }
+
+        /// <summary>
+        /// An object to hold and provide the scripts needed for the embedded scripting portion of the program.
+        /// </summary>
+        public static ScriptHandler Scripts = new ScriptHandler();
```
klaki892: It bothers me that this is a global variable that gets directly mentioned in the backend (ShellItemList)

We have `ConfigParser` and ConfigParser is almost visible by `ShellItemList` (Right now only ShellBagParser can see it) We handle all other configuration needed for the parsing operation to begin and i believe the flow of information of passing the Scripthandler to the ShellItemList (during construction?) would make the backend process not have a dependency on the frontend creation of the application itself (although this does change signatures of multiple classes),

Im surprised this isnt an issue that prevents the unit tests from running due to the unit tests not necessarily going through the Application context when running tests.

PR #156, opened 2020-02-18
HTML Output

### LINKED ISSUE #39: WPF HTML output
As an investigator or regular user, I want to view the ShellBag data in an HTML file with the graphical timeline.

bridCquinn:
This creates the HTML File of the filtered timeline. This also includes the Unit Test for it. Resolves #39

### DIFF
--- a/WPF/SeeShells/SeeShells/IO/HtmlIO.cs
+++ b/WPF/SeeShells/SeeShells/IO/HtmlIO.cs
@@ -1,0 +1,73 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.IO;
+using SeeShells.ShellParser.ShellItems;
+using SeeShells.UI;
+
+namespace SeeShells.IO
+{
+    public static class HtmlIO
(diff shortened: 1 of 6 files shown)


WPF/SeeShells/SeeShells/IO/HtmlIO.cs
```diff
@@ -0,0 +1,73 @@
...
+            outputFile.WriteLine(".grid-item {padding: 10px;border-left: 1px solid rgba(0, 0, 0, 0.8);text-align: center;}");
+            outputFile.WriteLine("</style>");
+            outputFile.WriteLine("</head>");
+
+            // Adding the shellItems to the body of thr html
+            outputFile.WriteLine("<body>");
+            outputFile.WriteLine("<h2>Timeline HTML Output</h2>");
+
+            // Iterates through the list of IEvents and adds the information to the HTML file
+            foreach (IEvent aEvent in eventList)
```
szczura: Do we need to be looping through a node list instead of an event list? Since we filter the nodes & we want to get the filtered list, we'd need that instead of the events, right? I may just be confused.

Issue #137, opened 2020-02-12
Default configurations for parsing 


klaki892:
As of #129, the user cannot use the program while offline due to the additional need to provide external files.
In the event that the program is a standalone exe without internet (e.g.  user downloads the program, transfers to a flashdrive, then uses the program on an offline computer)

We need to have defaults embedded in the program rather than throwing errors on no files found. 

Acceptance Criteria: 
- When clicking parse while there is no external files/invalid external files the program should use internal configurations and continue the parsing operation.  

_Originally posted by @klaki892 in https://github.com/RickLeinecker/SeeShells/pull/131_

PR #157, opened 2020-02-19
#140 Removes requirement of a script.json to parse

### LINKED ISSUE #140: Make scripts file optional
User doesn't NEED the scripts file; they should be able to parse without it.
Modify the home page to allow this.

klaki892:
- removes expired TODO and FIXME in home view page

This issue is a subset of #137 for being able to do parsing offline

Resolves #140


klaki892:
#146 has to be updated to safely handle not having any valid file passed to it in the event that this PR gets merged first.

PR #163, opened 2020-02-19
Connect the CsvIO to UI checkbox

### LINKED ISSUE #127: Allow user to export the csv
User should be able to open the application and export a CSV of their shellbag data.

Basically, do all the buttons and workflow involved with that.

Using online hive -> click parse -> check csv export -> hit download button -> get the csv

bridCquinn:
Connects the CSV function to UI. Resolve #127

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
+++ b/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
@@ -344,14 +344,18 @@ namespace SeeShells.UI.Pages
         }
 
         /// <summary>
-        /// This checks when the download button is hit, whether the HTML checkbox is checked or not and calls the creation of HtmlOutput.
+        /// This checks when the download button is hit, whether the HTML and/or the CSV checkbox is checked or not and calls the creation of HtmlOutput.
         /// </summary>
         private void download_Click(object sender, RoutedEventArgs e)
         {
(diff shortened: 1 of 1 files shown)


WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -344,15 +344,19 @@ private void SetTimeSpanSliderControlTextBoxValue(double sliderValue)
...
+            if(htmlCheckBox.IsChecked ?? false)
             {
-                System.Windows.MessageBox.Show("helps");
                 HtmlIO.OutputHtmlFile(App.nodeCollection.nodeList, "timeline.html");
             }
+            if(csvCheckBox.IsChecked ?? false)
+            {
+
+                CsvIO.OutputCSVFile(App.ShellItems, "raw.csv");
+            }
```
klaki892: We really should ask the user where to save the file.

The same applies to the HTML output, but Unfortunately got by us in #156  (could make a new issue, but that could be addressed here as well.)

szczura: we can prob make that a separate issue just to split it up

szczura: made issue #166 to address this

PR #167, opened 2020-02-21
Add script API endpoint adds a script to the database

### LINKED ISSUE #81: Add script API Endpoint
To complete:
- Endpoint can be hit
- A Lua script can be added to the database

### LINKED ISSUE #82: Get scripts API Endpoint
To be complete:
- hit the endpoint
- grab the scripts in the database

szczura:
Resolves #81 
Edited:
Resolves #82 

The add scripts endpoint will only accept base 64 encoded strings. When a user inputs a script on the website, the script will need to be encoded before being sent.

The get scripts endpoint will return these encoded strings. When the website receives these, it will need to decode them before displaying them to the user.

### DIFF
--- a/api/databaseconnection.js
+++ b/api/databaseconnection.js
@@ -194,6 +194,38 @@ function GUIDDoesNotExist(guid) {
     });
 }
 
+function addScript(identifier, script) {
+    return new Promise(function (resolve, reject) {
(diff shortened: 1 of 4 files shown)


szczura:
> 
> 
> Can we encode the script in transit (and storage) so that special characters, spacing, and formatting is preserved? storing the binary representation of the text (e.g. base64) would probably help with this rather than storing raw text.

Updated the code & reflected changes in the PR description.

PR #172, opened 2020-02-22
Embedded Scripting

### LINKED ISSUE #79: Make Lua script to parse known shell item

### LINKED ISSUE #147: WPF Get Lua Scripts & Make scripts.json file
Part of the SeeShells.IO.Networking folder

### LINKED ISSUE #148: WPF Use scripts.json (or other selected file) to get the scripts for ScriptHandler

szczura:
Resolves #147 
The code added in the IO Networking folder covers this part. The API endpoint is hit, and the JSON object's base64 scripts are decoded when de-serialized the first time for making the configuration file. The other times the JSON is de-serialized (reading from this file), a decoded scripts pair class is used so that it doesn't attempt to decode the scripts again.

Resolves #148 
The code uses the actual scripts.json configuration file instead of hard-coded scripts I had earlier in the program to test it out.

Resolves #79 
Control Panel Shell item script worked when used through the configuration file (like it did when it was used in my little test function within the program earlier this week). With how the code is written now, you won't see the script used since I took out the check for the 0x71 identifier. I did this since the script was just a proof of concept and so that the existing control panel shell item class can be used properly.

### DIFF
--- a/WPF/SeeShells/SeeShells/ShellParser/ShellItems/ShellItemList.cs
+++ b/WPF/SeeShells/SeeShells/ShellParser/ShellItems/ShellItemList.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using SeeShells.ShellParser.Scripting;
 using System.Collections.Generic;
 
 namespace SeeShells.ShellParser.ShellItems
@@ -28,9 +29,9 @@ namespace SeeShells.ShellParser.ShellItems
             { 
                 // if we have a script for the ShellItem, use it to get the information needed
                 int identifier = unpack_byte(off + 2);
-                if (App.Scripts.HasScriptForShellItem(identifier))
+                if (ScriptHandler.HasScriptForShellItem(identifier))
                 {
-                    return App.Scripts.ParseShellItem(buf, identifier);
+                    return ScriptHandler.ParseShellItem(buf, identifier);
                 }
 
                 // Getting here means that the ShellItem is unidentified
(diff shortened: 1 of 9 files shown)


WPF/SeeShells/SeeShells/ShellParser/ShellItems/ShellItemList.cs
```diff
@@ -24,13 +25,18 @@ protected IShellItem GetItem(int off)
...
             Type type = Type.GetType("SeeShells.ShellParser.ShellItems.ShellItem0x" + postfix);
-            if(type == null)
-            {
+            if (type == null)
+            { 
+                // if we have a script for the ShellItem, use it to get the information needed
+                int identifier = unpack_byte(off + 2);
+                if (ScriptHandler.HasScriptForShellItem(identifier))
+                {
+                    return ScriptHandler.ParseShellItem(buf, identifier);
```
klaki892: Because `ParseShellItem` (or rather `LuaShellItem`) can throw a multitude of exceptions now, we need to wrap this statement in a try/catch to prevent the entire program from crashing when a lua script fails to parse an ShellItem. an also give an error log message for the respective failure (e.g. why the script failed)

szczura: Did this in the latest push.

PR #185, opened 2020-02-24
Adds Offline username retrieval support

### LINKED ISSUE #164: Enable filtering Events by their owner
Currently we parse the entire registry and dont record metadata surrounding shell bag entries. One of the pieces of information we need for later filtering is which _user_ each Shell Bag (and subsequent Shell Items & events) came from.

Acceptance Criteria:
- User should be able to filter the timeline by the "owner" for the events displayed.

Work for this started in #55

klaki892:
- Fixes Online username error of setting last username to all values
- temporarily adds hardcoded username retrieval value which needs to be expanded upon

Resolves #164 However new issues/improvements need to be made. 
Specifically:
- Offline username retrieval is currently done through a hardcoded value which will only work on vista+ this value/potential keys that contain the username need to be added to a configuration.
- online username currently retrieve SID's because they are guaranteed and its non-trivial to resolve SID's to usernames without the SAM registry hive. Therefore live parsing should use the same username locations as offline and fallback to SID's when not available.

### DIFF
--- a/WPF/SeeShells/SeeShells/ShellParser/ConfigParser.cs
+++ b/WPF/SeeShells/SeeShells/ShellParser/ConfigParser.cs
@@ -82,7 +82,16 @@ namespace SeeShells.ShellParser
 
             return locations;
         }
-        
+
+        public List<string> GetUsernameLocations()
+        {
+            //todo retrieve from OSRegistryFile instead of hardcoded path
+            List<string> list = new List<string>();
+            // found username retrievalable key-value @ https://stackoverflow.com/a/53585223
+            list.Add(@"Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders");
+            return list;
+        }
+
         private string getLiveOSVersion()
         {
             RegistryKey registryKey = Microsoft.Win32.Registry.LocalMachine.OpenSubKey("Software\\Microsoft\\Windows NT\\CurrentVersion");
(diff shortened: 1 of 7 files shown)


WPF/SeeShells/SeeShells/ShellParser/ConfigParser.cs
```diff
@@ -82,7 +82,16 @@ public List<string> GetRegistryLocations()
...
             return locations;
         }
-        
+
+        public List<string> GetUsernameLocations()
+        {
+            //todo retrieve from OSRegistryFile instead of hardcoded path
+            List<string> list = new List<string>();
+            // found username retrievalable key-value @ https://stackoverflow.com/a/53585223
+            list.Add(@"Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders");
```
szczura: What other folders/keys can we find this info in?

klaki892: > What other folders/keys can we find this info in?

Because Shell Folders is a consistent created key across windows versions, this appears to be the only reliable registry key / values that include paths with the username. 

All other keys/values i found with a username value were circumstancial to what was already on the machine (recent searches, java is installed, etc.)

klaki892:
> Is there a reason why you did not write a unit test to check if you're retrieving the correct user ?

Because retrieving the correct user is specific to each implementation of ConfigParser, and username retrieval doesnt reveal results until the entire shell bag parsing is complete, its untestable other than our end-to-end tests for performing the entire process.

Before merging this issue i'll look to see if i can reasonably do it for our offline end-to-end parsing test.

klaki892:
> Is there a reason why you did not write a unit test to check if you're retrieving the correct user ?

Added a check to the offline test

PR #186, opened 2020-02-25
Timeline Remastered

### LINKED ISSUE #152: Dynamically generate the rectangles with time labels for the timelines
It needs to be scalable.

### LINKED ISSUE #165: Making one timeline out of multiple timelines
Use the timeline element in xaml and create multiple timelines that start after large chunks of time.

AlekStoyanov:
Events are clustered in 1 minute intervals of time, which are individual timelines.
Each timeline start time is rounded down to the nearest minute and each timeline end time is 1 minute from the start time.
The time stamps are placed on a separate stack panel due to stacking issues.
The events are still stacking since this is a later issue to be fixed by adding a different UI element to represent a stack of events.
None of the color choices are final and the whole timeline needs to be made prettier.
However, this is the first (ugly) iteration of timeline 2.0(Remastered).
Resolves #165 
Resolves #152


WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -239,134 +209,60 @@ public void BuildTimeline()
...
         {
-            if(htmlCheckBox.IsChecked ?? false)
-            {
-                SaveFileDialog saveFileDialog1 = new SaveFileDialog();
-                saveFileDialog1.DefaultExt = ".html";
-                saveFileDialog1.Filter = "Html File (*.html)| *.html";
-                saveFileDialog1.AddExtension = true;
-                saveFileDialog1.ShowDialog();
-                string name = saveFileDialog1.FileName;
-                HtmlIO.OutputHtmlFile(App.nodeCollection.nodeList, name);
```
szczura: Was all this accidentally deleted?

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -209,25 +151,53 @@ public void BuildTimeline()
...
+                List<Node.Node> nodesCluster = new List<Node.Node>(); // Holds events for one timeline at a time.
+                nodesCluster.Add(App.nodeCollection.nodeList[0]);
+                DateTime previousDate = App.nodeCollection.nodeList[0].aEvent.EventTime;
 
-                foreach (Node.Node node in App.nodeCollection.nodeList)
+                int nodeListSize = App.nodeCollection.nodeList.Count;
+                for (int i = 1; i < nodeListSize; i++)
+                {
+                    long ticks = previousDate.Ticks / maxRealTimeSpan.Ticks;
+                    DateTime realTimeStart = new DateTime(ticks * maxRealTimeSpan.Ticks); // This is the time of the first event rounded down to the nearest maxRealTimeSpan
```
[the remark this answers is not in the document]
AlekStoyanov: maxRealTimeSpan is a `TimeSpan` while previousDate is a `DateTime` so ticks are used in order to have one unified unit of measurement between them. 
As to why the math is twisted (divided then multiplied), I am not sure exactly, but this is how they show to do it here so I just took their word for it, and worked.
https://stackoverflow.com/questions/1393696/rounding-datetime-objects

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -209,25 +151,53 @@ public void BuildTimeline()
...
+
+                    // If the event belongs to the timeline
+                    if (App.nodeCollection.nodeList[i].aEvent.EventTime.Subtract(realTimeStart) < maxRealTimeSpan)
```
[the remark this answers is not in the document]
AlekStoyanov: The fist event is grabbed and it's time rounded down to make the start of the first timeline, then events are grabbed as long as they belong to that same minute of real time. The end of that first timeline is determined by the time of the last event rounded up. And that is done for all timelines. Meaning that not every timeline is exactly 1 minute long. In fact, most are shorter than that. That is kinda stupid, but that is what we discussed. I could make it be 1 minute exact for each timeline (which will add a little bit of a bigger gap of course, but at least all timelines will be equal). Or leave it as is and add end time in the block underneath rather than just a start time leaving the timelines unequal by some seconds of time.
We'll have to make a decision.
Good catch.

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
AlekStoyanov: That is very strange, I do not recall deleting this. I have never worked on this method, but I have somehow changed the comment and added code??? I merged masted into my branch before pushing, idk what this git black magic that is.
Do you have a clue ?

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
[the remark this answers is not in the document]
AlekStoyanov: The first event time is rounded down to the nearest minute to set the start of the first timeline. Then events are grabbed and placed on that timeline if they belong in the same minute interval. The first event that breaks that minute interval, is placed on another timeline with it's date rounded down to mark the begging of its timeline.

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
AlekStoyanov: It was. Good eye.

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -239,134 +209,60 @@ public void BuildTimeline()
...
-            {
-                if (minDate > node.aEvent.EventTime)
-                {
-                    minDate = node.aEvent.EventTime;
-                }
-            }
-            return minDate;
-        }
+            long ticks = nodesCluster[0].aEvent.EventTime.Ticks / maxRealTimeSpan.Ticks;
+            DateTime beginDate = new DateTime(ticks * maxRealTimeSpan.Ticks);
```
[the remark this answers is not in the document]
szczura: I understand this now.

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -147,87 +140,48 @@ private void EventNameFilter_TextChanged(object sender, TextChangedEventArgs e)
...
-                    Nodeline.Children.Add(node.dot);
+                    // If the event belongs to the timeline
+                    if (TimeSpan.Compare(App.nodeCollection.nodeList[i].aEvent.EventTime.Subtract(realTimeStart), maxRealTimeSpan) == -1) // Compare returns -1 if the first argument is less than the second
+                    {
+                        nodesCluster.Add(App.nodeCollection.nodeList[i]);
+                    }
+                    else
+                    {
+                        AddTimeline(nodesCluster);
+                        nodesCluster.Clear();
```
klaki892: How are we handling hidden events with this? wont this make empty timelines if a node in a timeline is hidden, but the timeline will still be made and reserved?

AlekStoyanov: Yes, you are right. I did not worry about filtering for this PR since I wanted to make a good foundation of what the timeline will look like after we are all happy with it and merge it, I was going to make another issue and add that functionality.
However, how are we planning on rebuilding the timeline after a filter is applied? Are you going to call the `BuildTimeline` method at the end of every filter method ?

klaki892: > Yes, you are right. I did not worry about filtering for this PR since I wanted to make a good foundation of what the timeline will look like after we are all happy with it and merge it, I was going to make another issue and add that functionality.
> However, how are we planning on rebuilding the timeline after a filter is applied? Are you going to call the `BuildTimeline` method at the end of every filter method ?

I assume this is the job of `NodeCollection` since it currently handles the management of filters. It would have to rebuild the timeline anytime a change in the states of the nodes would happen. (this shouldnt be done inside the individual filters)

PR #195, opened 2020-02-29
Filterable timeline

### LINKED ISSUE #194: Filterable Timeline
No timeline should be build if events are not visible (filtered).
Each timeline has to be scaled only according to visible events.
The timeline(s) have to be rebuilt every time a filter is applied, or removed.

AlekStoyanov:
The timeline is rebuilt every time a filter is applied or removed.
The rebuilt timeline only contains events that have not been filtered out and scales according to those events. 
Also added a check to the parse button that checks if a timeline has been built previously, and it it has it rebuilds it.

Resolves #194.

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Node/NodeCollection.cs
+++ b/WPF/SeeShells/SeeShells/UI/Node/NodeCollection.cs
@@ -1,9 +1,6 @@
 ﻿using SeeShells.UI.EventFilters;
-using System;
+using SeeShells.UI.Pages;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace SeeShells.UI.Node
 {
@@ -28,6 +25,7 @@ namespace SeeShells.UI.Node
         {
             filterList.Add(identifier, filter);
             filter.Apply(ref nodeList);
+            TimelinePage.timelinePage.RebuildTimeline();
         }
 
         /// <summary>
@@ -51,6 +49,7 @@ namespace SeeShells.UI.Node
                 {
                     filter.Apply(ref nodeList);
                 }
+                TimelinePage.timelinePage.RebuildTimeline();
             }
         }
 
--- a/WPF/SeeShells/SeeShells/UI/Node/NodeCollection.cs
+++ b/WPF/SeeShells/SeeShells/UI/Node/NodeCollection.cs
@@ -25,7 +25,7 @@ namespace SeeShells.UI.Node
         {
             filterList.Add(identifier, filter);
             filter.Apply(ref nodeList);
-            TimelinePage.timelinePage.RebuildTimeline();
+            Home.timelinePage.RebuildTimeline();
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -25,11 +26,12 @@ public partial class TimelinePage : Page
 
         private TimeSpan maxRealTimeSpan = new TimeSpan(0, 0, 1, 0); // Max time in one timeline (1 min).
 
+        public static TimelinePage timelinePage;
         public TimelinePage()
         {
             InitializeComponent();
-
             BuildTimeline();
+            timelinePage = this;
```
AlekStoyanov: I am not sure if capturing this object in a static member variable within itself in order to access it from other classes is the best practice, but it works and I can't think of any negative setbacks from it. If anyone has a better idea or any concerns, let me know.

szczura: I think instead of making this a static member, we could make a variable in the home page for the timeline page. Instead of .Navigate(new TimelinePage()), we can set a variable timelinePage to new TimelinePage() when it's not null and use .Navigate(timelinePage). This would keep the variable confined to one file instead of spreading it across 2.

AlekStoyanov: I like this idea, and I don't know why I did not think of it.

WPF/SeeShells/SeeShells/UI/Pages/Home.xaml.cs
```diff
@@ -282,7 +284,17 @@ private async void ParseButton_Click(object sender, RoutedEventArgs e)
...
-            NavigationService.Navigate(new TimelinePage());
+            if(timelinePage == null)
+            {
+                timelinePage = new TimelinePage();
+                NavigationService.Navigate(timelinePage);
+            }
+            else
+            {
+                timelinePage.RebuildTimeline();
+                NavigationService.Navigate(timelinePage);
```
AlekStoyanov: @klaki892 Changes will need to be made on the timeline page to reset all filters once parsing is done for a second time.

WPF/SeeShells/SeeShells/UI/Node/NodeCollection.cs
```diff
@@ -28,6 +25,7 @@ public void AddEventFilter(string identifier, INodeFilter filter)
         {
             filterList.Add(identifier, filter);
             filter.Apply(ref nodeList);
+            Home.timelinePage.RebuildTimeline();
```
szczura: I feel like a better place for this to live would be in TimelinePage.xaml.cs where the UpdateFilter function is.

This would keep the timeline UI display functionality confined in the TimelinePage file.

WPF/SeeShells/SeeShells/UI/Node/NodeCollection.cs
```diff
@@ -51,6 +49,7 @@ public void RemoveEventFilter(string filterIdentifer)
                 {
                     filter.Apply(ref nodeList);
                 }
+                Home.timelinePage.RebuildTimeline();
```
szczura: See my above comment.

WPF/SeeShells/SeeShells/UI/Pages/Home.xaml.cs
klaki892: Yeah, make an issue for it

PR #200, opened 2020-03-02
Website rework

### LINKED ISSUE #121: Fix website CSS linking

szczura:
Our website code is moved to the website folder. On pull requests, yarn build is done and the created files are put into the docs folder, committed, and pushed so that GitHub pages can pull from it.
Previously, GitHub pages was just looking at the index.html that was taken out of the public folder Vue builds would put it in. This is not how we should have been dealing with the files, and that was what was causing us problems.

Resolves #121 since we are starting from scratch and deploying with Vue properly.

I recommend not using/working from the template in the future since it seems to be more effort than it's worth. Bootstrapvue has some nice components we can use.

image


szczura:
#202 remade to deal with check failure

PR #202, opened 2020-03-02
Website rework

### LINKED ISSUE #121: Fix website CSS linking

szczura:
Our website code is moved to the website folder. On pull requests, yarn build is done and the created files are put into the docs folder, committed, and pushed so that GitHub pages can pull from it.
Previously, GitHub pages was just looking at the index.html that was taken out of the public folder Vue builds would put it in. This is not how we should have been dealing with the files, and that was what was causing us problems.

Resolves #121 since we are starting from scratch and deploying with Vue properly.

I recommend not using/working from the template in the future since it seems to be more effort than it's worth. Bootstrapvue has some nice components we can use.

image


szczura:
Added a commit that incorporates vue-router to make navigation simple.

Home currently looks like: 
image

and when you click on Administrative -> Register, you get:
image

.github/workflows/WebsiteBuildTest.yml
```diff
@@ -28,19 +28,35 @@ jobs:
...
       env:
         CI: true
+        
+    - name: Commit files
+      run: |
+        cd $GITHUB_WORKSPACE
+        git add docs/
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
+        git commit -m "Add docs folder with the build files" -a
```
klaki892: Although its by no means required, is it possible to reference the commit number from which this build was made?
That'll make the commits distinguishable by anyone who goes back and looks at the tree after everything is merged. (If we arent going to squash all future website commits from here on out)

website/src/App.vue
```diff
@@ -0,0 +1,31 @@
...
+        <NavigationBar />
+
+        <div id="page">
+            <router-view />
+        </div>
+    </div>
+</template>
+
+<script>
+    import NavigationBar from './components/NavigationBar.vue'
```
klaki892: Is it okay this isnt at the top of the file?

.github/workflows/WebsiteBuildTest.yml
szczura: Should be doable I think.

website/src/App.vue
szczura: Yup. It seems like most vue components are structured html -> javascript -> css

PR #205, opened 2020-03-04
Clickable objects

### LINKED ISSUE #87: Create clickable object
Display details about the shell item events when hovering over a graphical representation of it on the timeline.

bridCquinn:
InformedDot was created to make a clickable object which references the event which it was built for. During some merge conflicts the circle portion of the dot was lost, but that is aesthetics and will be dealt with in another issue. Right now the method run when a dot is clicked is a place holder that shows a MessageBox of information about the Shell Item. Reslove #87


bridCquinn:
The new pushes to this branch were made to to fix the null reference that kept failing the nodecollectiontest. 
The current structure was implemented and kept because System.Windows.Controls cannot be in a class which is then extended by another. This was happening in MockNode, which meant that it wouldn't even build, the workaround for this problem is basically making an intermediary class which doesn't extend that class, but holds the properties needed. In our case, this was pretty much what I did with InformedDot.

bridCquinn:
Resolve #87

WPF/SeeShells/SeeShells/UI/Node/NodeParser.cs
```diff
@@ -36,22 +40,31 @@ public static List<Node> GetNodes(List<IEvent> eventList)
...
         /// <summary>
         /// Sets up initial properties for a graphical block object
         /// </summary>
         /// <param name="block">graphical object that contains event details on a timeline</param>
-        private static void SetBlockProperties(Rectangle block)
+        private static void SetBlockProperties(TextBlock block, IEvent aEvent)
         {
-            /// TODO:
-            /// Set Properties.
+            //foreach (KeyValuePair<string, string> property in aEvent.Parent.GetAllProperties())
```
klaki892: Why are we no longer pulling all properties?

bridCquinn: This was done because as of how the test runs with a null parent. So I made a few other issues and the issue about hovering objects will have to address that. As of now this only shows the 3 properties found in event. In short, to make the test run properly.

klaki892: > This was done because as of how the test runs with a null parent. So I made a few other issues and the issue about hovering objects will have to address that. As of now this only shows the 3 properties found in event. In short, to make the test run properly.

We shouldn't be breaking functionality to conform to tests. Tests are there to make sure the functionality is properly working, there's no rule against changing a test to make sure it still meets the defined parameters. 
Correct me if I'm wrong, but if the issue is a null parent in the event, why can't you set the parent in the test event for the test case? 
It's alright if the functionality is being moved to a separate section, but I'm not sure if that's what's going on here. (e.g. The functionality of displaying all information is moved to a different section.)
If anything change the test case to maintain our functionality, or add the requirement of getting all properties visible as a condition of completing #214

bridCquinn: The functionality of showing all properties is under the hover issue.

PR #211, opened 2020-03-06
Special character support for CSV Import/Export

### LINKED ISSUE #206: CSVExport cant handle special characters inside shellbag data
It was found that a line break was created while exporting CSV data from Shellbags from a live registry.

This leads to the greater realization that special characters which are parsed from binary data cannot be handled by our current CSV parser from exisitng `IShellItems` 

Acceptance Criteria:
- The CSV Exporter must be modified so that special characters are handled correctly and can successfully be read back by the CSV Importer.

AlekStoyanov:
Export:
Encloses every cell in double quotes in order to preserve all special characters in that cell.
If there are double quotes in a given cell, they get wrapped in double quotes.
For more info: https://hawq.apache.org/docs/userguide/2.3.0.0-incubating/datamgmt/load/g-escaping-in-csv-formatted-files.html

Import:
Uses CSVHelper (imported from NuGet) to parse a CSV file that is being imported.

TestCase:
Exports a CSV for a shellitem that contains special characters in the name field, and parses that file to check if all fields were read correctly into memory.

*Note: 
@yara-58, you should pull this branch and try exporting and then importing your ShellBags on the laptop you used when you found this bug, to further confirm that the bug is fixed.

Resolves #206

### DIFF
--- a/WPF/SeeShells/SeeShells/packages.config
+++ b/WPF/SeeShells/SeeShells/packages.config
@@ -1,11 +1,16 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="CsvHelper" version="15.0.1" targetFramework="net46" />
   <package id="Extended.Wpf.Toolkit" version="3.8.1" targetFramework="net46" />
   <package id="KeraLua" version="1.0.24" targetFramework="net46" />
+  <package id="Microsoft.CSharp" version="4.5.0" targetFramework="net46" />
   <package id="Newtonsoft.Json" version="12.0.3" targetFramework="net46" />
   <package id="NLog" version="4.6.8" targetFramework="net46" />
   <package id="NLog.Config" version="4.6.8" targetFramework="net46" />
   <package id="NLog.Schema" version="4.6.8" targetFramework="net46" />
   <package id="NLua" version="1.4.27" targetFramework="net46" />
   <package id="RestSharp" version="106.10.1" targetFramework="net46" />
+  <package id="System.Runtime.CompilerServices.Unsafe" version="4.5.2" targetFramework="net46" />
+  <package id="System.Threading.Tasks.Extensions" version="4.5.2" targetFramework="net46" />
+  <package id="System.ValueTuple" version="4.4.0" targetFramework="net46" />
 </packages>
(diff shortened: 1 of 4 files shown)


WPF/SeeShells/SeeShellsTests/IO/CsvIOTests.cs
```diff
@@ -47,5 +47,43 @@ public void ImportCSVFileTest()
...
+        [TestMethod()]
+        public void ExportAndImportCSVWithSpecialCharactersTest()
+        {
+            // Export
+            List<IShellItem> shellItems = new List<IShellItem>();
+            Dictionary<string, string> shellItemProperties = new Dictionary<string, string>();
+            shellItemProperties.Add("Size", "0");
+            shellItemProperties.Add("Type", "31");
+            shellItemProperties.Add("TypeName", "Some Type Name");
+            shellItemProperties.Add("Name", "Some Name, \n \"Name\"");
```
klaki892: Because binary data could be read into a non dispalyable character, we might need to test more than just a newline character.

AlekStoyanov: Since I am enclosing each cell in double quotes when exporting, any character inside a cell is escaped. Therefore, no matter what character we have inside, it will not affect the CSV formatting.

WPF/SeeShells/SeeShells/packages.config
```diff
@@ -1,11 +1,16 @@
...
   <package id="Extended.Wpf.Toolkit" version="3.8.1" targetFramework="net46" />
   <package id="KeraLua" version="1.0.24" targetFramework="net46" />
+  <package id="Microsoft.CSharp" version="4.5.0" targetFramework="net46" />
   <package id="Newtonsoft.Json" version="12.0.3" targetFramework="net46" />
   <package id="NLog" version="4.6.8" targetFramework="net46" />
   <package id="NLog.Config" version="4.6.8" targetFramework="net46" />
   <package id="NLog.Schema" version="4.6.8" targetFramework="net46" />
   <package id="NLua" version="1.4.27" targetFramework="net46" />
   <package id="RestSharp" version="106.10.1" targetFramework="net46" />
+  <package id="System.Runtime.CompilerServices.Unsafe" version="4.5.2" targetFramework="net46" />
```
szczura: What is this for?

AlekStoyanov: Probably something related to CsvHelper.

PR #216, opened 2020-03-07
Node fixes

### LINKED ISSUE #87: Create clickable object

bridCquinn:
This has the same effect as the old pull request to resolve clickable objects, but without an extra class. Resolve #87

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Node/Node.cs
+++ b/WPF/SeeShells/SeeShells/UI/Node/Node.cs
@@ -4,14 +4,15 @@ using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using System.Windows.Shapes;
+using System.Windows.Controls;
+using System.Windows;
 
 namespace SeeShells.UI.Node
 {
-    public class Node
+    public class Node : Button
     {
         public IEvent aEvent;
-        public Ellipse dot;
-        Rectangle block;
+        public TextBlock block;
 
         /// <summary>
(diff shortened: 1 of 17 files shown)


WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -304,5 +304,13 @@ private void download_Click(object sender, RoutedEventArgs e)
                 CsvIO.OutputCSVFile(App.ShellItems, name2);
             }
         }
+
+        /// <summary>
+        /// This activates the toggle_block method built into the InformedDot object. 
+        /// </summary>
```
AlekStoyanov: Since we are no longer using InformedDot, this comment is stale.

WPF/SeeShells/SeeShells/UI/Node/Node.cs
```diff
@@ -4,26 +4,41 @@
...
-            this.dot = dot;
             this.block = block;
         }
+
+        /// <summary>
+        /// This is used to hide and show the block of information connected to each dot of information on the timeline.
+        /// </summary>
+        public void toggle_block()
+        {
+            MessageBox.Show(block.Text);
```
klaki892: Correct me if I'm wrong, but won't a message box block input to the rest of the program? Making it impossible for a user to see detailed information for more than one event?

PR #226, opened 2020-03-14
Pretty timeline

### LINKED ISSUE #225: Make the timeline more visually intuitive
Redesign the timestamps of the current timeline to make the timeline more intuitive:
1. Add ticks to graph out each second of the timeline.
2. Add separation elements to represent gaps of time in the timeline.
3. Add lines that stretch from each node to it's exact position on the timeline.

AlekStoyanov:
1. Adds ticks to graph out each second of the timeline. (the ticks are placed in a TimelinePanel which handles their spacing)
2. Adds separation lines to designate the end of one timeline and start of another.
3. Adds a timeline bar to represent the continuous timeline.
4. Adds lines to connect each node to the timeline bar in order to pinpoint their location in relation to the ticks on the bar.

Note:
Due to a significant amount of objects being added to the UI, the program has a slight slowdown. It takes about extra 3 to 4 seconds to parse and 0 to 4 seconds to redraw after filters are applied depending on how large the filtered amount of events is. However, I am running it on pretty fast hardware, thus the slowdown might be more severe on lower end machines. This will be something to test and possibly optimize in the future.

image

Resolves #225

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Node/NodeParser.cs
+++ b/WPF/SeeShells/SeeShells/UI/Node/NodeParser.cs
@@ -39,8 +39,8 @@ namespace SeeShells.UI.Node
         /// <param name="node">Acta as a graphical object (button) to be placed on the timeline.</param>
         private static void SetNodeProperties(Node node)
         {
-            node.Width = 10;
-            node.Height = 10;
+            node.Width = 20;
+            node.Height = 20;
             node.Click += Pages.TimelinePage.Dot_Press;
         }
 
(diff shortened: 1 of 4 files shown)


AlekStoyanov:
I could make the nodes bigger. The only downside is that one whole timeline (1 minute of time) won't fit on the screen at once since the gaps left for each second will be bigger. I'll make them bigger and show you screenshots.

About the slowness, it is not due to the filtering, it is due to the drawing of all the elements on the UI.

PR #232, opened 2020-03-18
Updating scripts on the website

### LINKED ISSUE #23: Website update or add a Lua script
As a maintainer of SeeShells, I want to update the database with new shell item scripts.

szczura:
Resolves #23 

Also removed the use of passportJS in the API since it wasn't working out well. It's intended more for applications that render server-side.

When logged in, a user can visit this page:
image

If any scripts already exist, it will allow them to edit the existing ones:
image

There is some explanation on how the script works and what functions are available for use from the LuaShellItem class:
image

When you click on the functions, a better explanation expands:
image


api/index.js
```diff
@@ -367,28 +402,56 @@ app.get('/getHelpInformation', function (req, res) {
...
 
 
 app.get('/logout', function (req, res) {
-    req.logout();
-    res.redirect('/');
+    let promise = database.getSession(req.header('x-auth-token'));
+    promise.then(
+        function (result) {
+            if (result == true) {
+                let promise = database.destroySession(req.header('x-auth-token'));
```
klaki892: If someone doesn't explicitly logout, and the session token expires, will we still have a stale session in the database forever?

api/index.js
```diff
@@ -25,77 +24,25 @@ app.use(database.session({
     secret: process.env.SESSION_SECRET,
     resave: false,
     saveUninitialized: false,
-    cookie: { maxAge: 30 * 60 * 1000 }, // 30 minutes
+    cookie: { maxAge: 30 * 60 * 1000, secure:false }, // 30 minutes
```
klaki892: why secure:false?

api/index.js
szczura: The middleware I am using will prune any expired sessions. The default for session pruning is doing it every 60 seconds. I have left sessions in the database without deleting them to see if this will actually work, and the sessions do get deleted once their expiration time passes. 

More info here: https://www.npmjs.com/package/connect-pg-simple

api/index.js
szczura: I had this while doing development because it was giving me difficulty with figuring out what was going on with the session data I was sending. (Some info here:https://stackoverflow.com/questions/40324121/express-session-secure-true and here: https://stackoverflow.com/questions/40324121/express-session-secure-true )

Now that everything seems to be working fine, I think I can set it back to true.

Issue #201, opened 2020-03-02
Timeline Children Assignment Crashes WPF 


AlekStoyanov:
Secondary parsing causes the WPF to crash due to timeline children elements not being cleared properly.

Issue #189, opened 2020-02-26
Integrate Timeline/Help Navigation into Toolbar 


klaki892:
(dependent on completion of #187) 

Currently navigation between the timeline and help page happen in tabular format. to provide a cleaner user experince this should be changed so that its possible to navigate between the two pages (and the home page) by using the options in a dropdown menu.

Conditional access to the views may be needed to ensure smooth user experience.
(e.g. not accessing the timeline page until a parse has occurred at least once)

Acceptance Criteria:
- Modify and provide functionality into the toolbar to provide working functionality for the "View" Dropdown:
   - Home Page
   - Timeline Page
   - Help Page
- Removal of the back and forward buttons for navigating between pages
- Removal of the Tabs on top of the timeline and help page views.

PR #233, opened 2020-03-19
Homeand timeline navigation

yara-58:


WPF/SeeShells/SeeShells/App.xaml
```diff
@@ -45,8 +45,8 @@
                                     RelativeSource={RelativeSource FindAncestor, AncestorType={x:Type WrapPanel}}}" Background="Transparent">
                                     <Menu Background="Transparent">
                                         <MenuItem Header="View" Foreground="DarkGray" >
-                                            <MenuItem Header="Home"/>
-                                            <MenuItem Header="Timeline"/>
+                                            <Hyperlink NavigateUri="/SeeShells;component/ui/pages/home.xaml">Home</Hyperlink>
+                                            <Hyperlink NavigateUri="/SeeShells;component/ui/pages/timelinepage.xaml">Timeline Page</Hyperlink>
```
klaki892: I am concerned this might directly conflict with one of the acceptance criteria of #189 :
> Removal of the back and forward buttons for navigating between pages

When testing this currently, clicking on Home or Timeline Page from the toolbar generate's these two buttons on the UI.

klaki892: **@Team,** Do we want to prevent the user from being able to navigate to the timeline before at least one parsing action has taken place?

AlekStoyanov: I say yes.

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -29,7 +29,14 @@ public partial class TimelinePage : Page
...
             InitializeComponent();
-            BuildTimeline();
+            if (Home.timelinePage == null)
+            {
+                BuildTimeline();
+            }
+            else
+            {
+                Home.timelinePage.RebuildTimeline();
+            }
```
szczura: This whole section is bizarre. We're creating new Timeline pages while using the old original one? Am I misunderstanding this? I think this will create a memory issue where we keep creating pages but are not doing anything with them.

yara-58:
After soliciting Klaton's help the solution we came up with is that we had to make a global variable for NavigationService in App.xaml.cs because of the complexity of navigating within a wpf. The NavigationService only works if it is being referenced from a page. I exhausted every other option on this front and this is the only one that I found worked. Originally we had created a global variable for the TimelinePage in app.xaml as well but that was not the most ideal solution. I created a dictionary of pages instead. This will also allow us to add other pages we may want to navigate to in the future.

yara-58:
it does i just messed something up when i tried to resolve a merge conflict. trying to fix it

WPF/SeeShells/SeeShells/App.xaml
```diff
@@ -43,15 +44,7 @@
                             <ToolBarTray IsLocked="True" Background="Transparent">
                                 <ToolBar OverflowMode="Never" Loaded="Toolbar_Loaded" MinWidth="{Binding ActualWidth,
                                     RelativeSource={RelativeSource FindAncestor, AncestorType={x:Type WrapPanel}}}" Background="Transparent">
-                                    <Menu Background="Transparent">
-                                        <MenuItem Header="View" Foreground="DarkGray" >
-                                            <MenuItem Header="Home"/>
-                                            <MenuItem Header="Timeline"/>
-                                        </MenuItem>
-                                        <MenuItem Header="Help" Foreground="DarkGray">
-                                            <MenuItem Header="About SeeShells" Click="About_OnClick"/>
```
klaki892: This removes our ability to see the "About SeeShells" page.
Probably dont want to remove that

WPF/SeeShells/SeeShells/UI/Templates/Switch.xaml
```diff
@@ -0,0 +1,20 @@
...
+             mc:Ignorable="d" 
+             d:DesignHeight="450" d:DesignWidth="800">
+    <Grid>
+        <Menu Background="Transparent">
+            <MenuItem Header="View" Foreground="DarkGray" >
+                <Hyperlink NavigateUri="/SeeShells;component/ui/pages/home.xaml" Foreground="Black">Home</Hyperlink>
+                <MenuItem x:Name="timeline" Header="Timeline Page" Click="MenuItem_Click" Foreground="Black" />
+            </MenuItem>
+            <MenuItem Header="Help" Foreground="DarkGray" >
+                <Hyperlink NavigateUri="/SeeShells;component/ui/pages/helppage.xaml" Foreground="Black">About</Hyperlink>
```
klaki892: Currently if you click anywhere else in the button other than on the hyperlink text specifically, the button does nothing:

4d2301f1-3ed7-4b9d-9a62-e89cbe8fe7d3

Can we make it so the entire button is a hyperlink / goes to its destination?

klaki892: `About SeeShells` and `Help` are two separate things, leading to different parts of the program. There should probably be two items in the dropdown. 
```suggestion
                <Hyperlink NavigateUri="/SeeShells;component/ui/pages/helppage.xaml" Foreground="Black">Help</Hyperlink>
```

WPF/SeeShells/SeeShells/UI/Templates/Switch.xaml.cs
```diff
@@ -0,0 +1,50 @@
...
+            else
+            {
+                foreach(var el in App.pages)
+                {
+                    if (el.Key.Contains("timelinepage"))
+                    {
+                        App.NavigationService.Navigate(el.Value);
+
+                    }
+                }
```
klaki892: You dont need to loop over a dictionary, you can check for the key/value specifically.
```suggestion
                string timelinePageKey = "timelinepage";
                if (App.pages.ContainsKey(timelinePageKey))
                {
                    App.NavigationService.Navigate(App.pages[timelinePageKey]);
                }
```

WPF/SeeShells/SeeShells/UI/Templates/Switch.xaml
yara-58: I looked into it and there is something called a HyperlinkButton but the wpf is not able to find it as a control for some reason: https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.hyperlinkbutton 

As far as I can tell the hyperlink itself is attached only to the text

WPF/SeeShells/SeeShells/UI/Pages/Home.xaml.cs
```diff
@@ -296,13 +299,17 @@ private async void ParseButton_Click(object sender, RoutedEventArgs e)
...
             }
             else
             {
                 timelinePage.RebuildTimeline();
                 NavigationService.Navigate(timelinePage);
             }
-            
+
+            App.NavigationService = NavigationService;
+            App.pages.Add("timelinepage", timelinePage);
```
szczura: Check if timelinepage exists in the dictionary before we add it.

WPF/SeeShells/SeeShells/UI/Templates/Switch.xaml
```diff
@@ -12,8 +12,9 @@
                 <Hyperlink NavigateUri="/SeeShells;component/ui/pages/home.xaml" Foreground="Black">Home</Hyperlink>
                 <MenuItem x:Name="timeline" Header="Timeline Page" Click="MenuItem_Click" Foreground="Black" />
             </MenuItem>
-            <MenuItem Header="Help" Foreground="DarkGray" >
-                <Hyperlink NavigateUri="/SeeShells;component/ui/pages/helppage.xaml" Foreground="Black">About</Hyperlink>
+            <MenuItem Header="Help" Foreground="DarkGray">
+                <Hyperlink NavigateUri="/SeeShells;component/ui/pages/helppage.xaml" Foreground="Black" >Help</Hyperlink>
+                <Hyperlink Foreground="Black">About</Hyperlink>
```
klaki892: This does not restore the functionality to display the About Page, see my previous comment about the code that must be placed back.

klaki892: The previously mentioned issue about Buttons not being clickable is still present (only the text in the buttons are) 
This needs to be addressed.

WPF/SeeShells/SeeShells/UI/Templates/Switch.xaml
```diff
@@ -9,12 +9,12 @@
...
             <MenuItem Header="View" Foreground="DarkGray" >
-                <Hyperlink NavigateUri="/SeeShells;component/ui/pages/home.xaml" Foreground="Black">Home</Hyperlink>
+                <MenuItem Header="Home" Click="Home_Click" Foreground="Black"/>
                 <MenuItem x:Name="timeline" Header="Timeline Page" Click="MenuItem_Click" Foreground="Black" />
             </MenuItem>
             <MenuItem Header="Help" Foreground="DarkGray">
-                <Hyperlink NavigateUri="/SeeShells;component/ui/pages/helppage.xaml" Foreground="Black" >Help</Hyperlink>
-                <Hyperlink Foreground="Black">About</Hyperlink>
+                <MenuItem Header="Help page" Click="Help_Click" Foreground="Black"/>
+                <MenuItem Header="About" Foreground="Black"/>
```
klaki892: This _**Still**_ does not re-enable navigation to the the About Me window. 
Please look at the original overwritten code for re-enabling this functionality

yara-58:
The about me window has been restored and this pull requests also closes issue #201

PR #234, opened 2020-03-19
Import csv

yara-58:


WPF/SeeShells/SeeShells/UI/Templates/Switch.xaml.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+
+        private void Import_Click(object sender, RoutedEventArgs e)
+        {
+
+            if (Home.timelinePage == null)
+            {
+                csvimport.IsEnabled = false;
+            }
```
klaki892: Importing from CSV is not dependent on the user having performed a parse operation already.

Importing from a CSV should be able to be done _instead_ of a normal parse operation, so it saves the user time. 
Requiring that they parse over some data before they are allowed to import a previous "run" of the program is counterproductive

WPF/SeeShells/SeeShells/UI/Templates/Switch.xaml.cs
```diff
@@ -0,0 +1,97 @@
...
+                List<IEvent> events = EventParser.GetEvents(App.ShellItems);
+                App.nodeCollection.ClearAllFilters();
+                App.nodeCollection.nodeList.AddRange(NodeParser.GetNodes(events));
+
+                Home.timelinePage.RebuildTimeline();
+                string timelinePageKey = "timelinepage";
+                if (App.pages.ContainsKey(timelinePageKey))
+                {
+                    App.NavigationService.Navigate(App.pages[timelinePageKey]);
+                }
```
klaki892: Don't make changes that are requested on other branches/PR's onto another branch(in this case, this one). This will cause some nasty merge conflicts if you make the fixes on both branches. 
because this branch is dependent on #233, its probably best to rebase instead of merging when working on a dependent branch. this way you can keep only the changes that specifically affect this issue separate from other branches that have their own problems and resolutions, that way you can rebase as those other branches get updated/ merge into master and your fixes in this branch will still supersede previous edits.

To help undo the mess that will come from this, i suggest hard resetting before the `merge` commit. _Force push_ once you're done with the reset but **ONLY** when you're sure everything is where it should be and nothing will actually get deleted. 

Sorry, I might have been preemptive by telling you to merge/rebase so you wouldnt get blocked by the other issue being in an active PR. Keeping the changes separate is a bit tasking and requires actively switching between branches to put changes in the right place and if something goes wrong it turns into a cascade of errors.

WPF/SeeShells/SeeShells/UI/Templates/Switch.xaml.cs
yara-58: The things is we run into the same issue with navigating to the timeline page after they have imported their csv because we cant navigate to the timeline page until the home page navigates there. We could possibly navigate to the timeline page from the main window when the importcsv is clicked.

klaki892: From our recent discussion we know that `NavigationService`:
1. does not exist during creation of the page
2. does exist by the time we hit parse button

You should be able to get the `NavigationService `reference needed for the toolbar before the user has to do a parse operation with those known facts. look possibly into finding the reference during `OnLoad` or immediately right afterwards.

WPF/SeeShells/SeeShells/UI/Templates/Switch.xaml.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                string timelinePageKey = "timelinepage";
+                if (App.pages.ContainsKey(timelinePageKey))
+                {
+                    App.NavigationService.Navigate(App.pages[timelinePageKey]);
+                }
+            }
+        }
+
+        private void Import_Click(object sender, RoutedEventArgs e)
```
klaki892: Right now the UI Freezes with no possible interaction. This can be made into a new issue but the user must know the program is working during this time consuming operation.

yara-58: I can make this a new issue

WPF/SeeShells/SeeShells/UI/Templates/Switch.xaml.cs
```diff
@@ -67,5 +72,50 @@ private void About_OnClick(object sender, RoutedEventArgs e)
...
+                Home.timelinePage.RebuildTimeline();
+                string timelinePageKey = "timelinepage";
+                if (App.pages.ContainsKey(timelinePageKey))
+                {
+                    App.NavigationService.Navigate(App.pages[timelinePageKey]);
+                }
+                else
+                {
+                    App.NavigationService.Navigate(Home.timelinePage);
+                }
```
yara-58: There was a slight issue where if you imported a csv and it took you to the timeline page, when you navigate back to the home page and try to import another csv it wouldnt navigate to the timeline page but this line resolves that.

PR #235, opened 2020-03-19
Adds Support for getting Help content from API

### LINKED ISSUE #231: Obtain & Cache new Help pages on the WPF
As an initial requirement of the project, in order to support longevity without changing the binary of the application, the help page of the application should be able to update its content.

To meet this end, the Help section of the application has two basic requirements:
- Obtain from our API (#180) updated versions of help as they become available
    - (additionally) Cache the help for subsequent offline use of the program
- Provide initial offline help in the event the program never connects to the internet. 

To meet these ends, the Help page can either attempt to obtain updates automatically or request the user's permission to change the help contents.

Acceptance Criteria: 
- A Method for obtaining and displaying updated help content is implemented into the WPF.
- A method for saving/caching updated help content is present once updated help is obtained.
- Help content exists and is fully functional without internet access.

klaki892:
- Adds help content caching into current executable directory
- Refactored API slightly to decouple writing to FS & getting API data

f3b72f58-e948-4047-a52e-be5267d57106

Current testing within the current Help API specification, the markdown seems to work when we put just plain text into the database. Depending on how the website will want to render it, we might have to switch to encoding the help content itself (e.g. base64)

Resolves #231

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Pages/HelpPage.xaml
+++ b/WPF/SeeShells/SeeShells/UI/Pages/HelpPage.xaml
@@ -5,6 +5,7 @@
       xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
       xmlns:local="clr-namespace:SeeShells.UI.Pages"
       xmlns:wpf="clr-namespace:Markdig.Wpf;assembly=Markdig.Wpf"
+      xmlns:xctk="http://schemas.xceed.com/wpf/xaml/toolkit"
       mc:Ignorable="d" 
       d:DesignHeight="450" d:DesignWidth="800"
       Title="HelpPage"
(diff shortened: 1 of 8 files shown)


WPF/SeeShells/SeeShells/UI/Pages/HelpPage.xaml.cs
```diff
@@ -28,34 +29,77 @@ namespace SeeShells.UI.Pages
     /// </summary>
     public partial class HelpPage : Page
     {
+
+        private static NLog.Logger logger = NLog.LogManager.GetCurrentClassLogger();
+
         private const string ReadmeFile = "README.md";
```
klaki892: Im debating changing this to `Help.md` React :+1: if it should be renamed or  :-1: if it's not ambigious to anyone who will see the file after it gets created.

AlekStoyanov: Personally when I see README, in my mind it's supposed to be a text file you read that has release info and any installation instructions. But since this is a resource file to never be read from the user directly and rather only intended to be accessed by our WPF, I think it should be renamed to Help.md

WPF/SeeShells/SeeShells/UI/Pages/HelpPage.xaml
```diff
@@ -14,7 +15,8 @@
     </FrameworkElement.CommandBindings>
 
     <Grid Background="White">
-        <wpf:MarkdownViewer x:Name="HelpViewer" Markdown="[Configuration Reference](https://github.com/Kryptos-FR/markdig.wpf). ">
+        <xctk:BusyIndicator x:Name="LoadingIndicator"  Grid.ZIndex="1"  IsBusy="True" BusyContent="Checking for Updates..."/>
```
szczura: This is so nice.

PR #243, opened 2020-03-22
Clickable objects, hover objects, style

### LINKED ISSUE #213: Graphical Boxes of Information
At the moment, the information is connected properly to the nodes on the timeline, and can display it with messageboxes, but the the information in event (name, type, time) need to be displayed when a node is clicked and also the nodeclump is clicked.

### LINKED ISSUE #214: Hover on the boxes
After the graphical boxes are created and clicked, the user needs some was to see all the information in the shell Item (Node). Since having a constant large amount of information about one shell item is not necessary, then a hover will work, so that people can see the information, when it does away to not block other elements.

### LINKED ISSUE #215: Style node and node cluster
The easiest way to make a uniform look on the timeline and achieve the exact design we want, a style will need to be made for each type of button on the timeline.

bridCquinn:
This has the style for the textblock, node and structurednode. Resolve #215
Clicking Nodes works. Resolves #213 
Clicking structurednodes 
Hovering works on textblocks and reveals extra information. Resolves #214

### DIFF
--- a/WPF/SeeShells/SeeShells/App.xaml.cs
+++ b/WPF/SeeShells/SeeShells/App.xaml.cs
@@ -26,7 +26,7 @@ namespace SeeShells
         public static NodeCollection nodeCollection = new NodeCollection();
 
         /// <summary>
-        /// Collectin of <see cref=MockShellItem"/> which is populated after a parsing operation.
+        /// Collection of <see cref="ShellItem"/> which is populated after a parsing operation.
         /// </summary>
         public static List<IShellItem> ShellItems { get; set; }
 
--- a/WPF/SeeShells/SeeShells/App.xaml.cs
+++ b/WPF/SeeShells/SeeShells/App.xaml.cs
@@ -26,7 +26,7 @@ namespace SeeShells
         public static NodeCollection nodeCollection = new NodeCollection();
 
         /// <summary>
-        /// Collectin of <see cref="ShellItem"/> which is populated after a parsing operation.
+        /// Collectin of <see cref="MockShellItem"/> which is populated after a parsing operation.
         /// </summary>
         public static List<IShellItem> ShellItems { get; set; }
 
--- a/WPF/SeeShells/SeeShells/App.xaml.cs
+++ b/WPF/SeeShells/SeeShells/App.xaml.cs
@@ -26,7 +26,7 @@ namespace SeeShells
         public static NodeCollection nodeCollection = new NodeCollection();
 
         /// <summary>
-        /// Collectin of <see cref="MockShellItem"/> which is populated after a parsing operation.
+        /// Collectin of <see cref=MockShellItem"/> which is populated after a parsing operation.
         /// </summary>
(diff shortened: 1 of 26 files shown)


WPF/SeeShells/SeeShells/UI/Node/InfoBlock.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+            else
+            {
+                this.Width = 450;
+                this.Height = 250;
+                this.Text = "";
+                foreach (KeyValuePair<string, string> property in this.aEvent.Parent.GetAllProperties())
+                {
+                    this.Text += property.Key + "," + property.Value;
+                    this.Text += "\n";
```
klaki892: image

Can we change this to be more user-friendly?
Some ideas (can do any/all):

- [ ]  Change the comma to a colon and add spacing between the key and the value
- [ ]  Display the info in a grid? like | Key | Value |
- [ ]  (potentially dangerous) Right now we use camel casing as a backend representation, can we seperate the words in the `key` that do this by spaces?

bridCquinn: I don't mind doing this, but should it be a different issue?

klaki892: > I don't mind doing this, but should it be a different issue?

It definitely can be, this issue meets the requirement of just displaying the info. 
Enhancements can come later

bridCquinn: I did the quick change between comma and colon. I'll look into the 3rd suggestion, soon, but I won't fret about it for this PR.

WPF/SeeShells/SeeShells/App.xaml.cs
```diff
@@ -26,7 +26,7 @@ public partial class App : Application
         public static NodeCollection nodeCollection = new NodeCollection();
 
         /// <summary>
-        /// Collectin of <see cref="ShellItem"/> which is populated after a parsing operation.
+        /// Collectin of <see cref=MockShellItem"/> which is populated after a parsing operation.
```
klaki892:
```suggestion
        /// Collectin of <see cref="ShellItem"/> which is populated after a parsing operation.
```

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -142,6 +143,11 @@ private void EventNameFilter_TextChanged(object sender, TextChangedEventArgs e)
 
         }
 
+        public static void click_line(object sender, EventArgs e)
+        {
+            System.Windows.MessageBox.Show("This is the textblock");
```
klaki892: Should this still be here?

bridCquinn: It is not, I'll take it out.

PR #249, opened 2020-03-22
Adds buttons to clear each individual filter

### LINKED ISSUE #203: Clear filters button/toolbar option
It's not super intuitive how to remove some of the filters like filtering by dates (need to click on them and backspace the dates).
Having a button of some kind for users to click to remove filters would be nice.

AlekStoyanov:
Adds a clear button to each individual filter.
I did not add a clear all filter as it seems unnecessary.

image

Resolves #203

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml
+++ b/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml
@@ -38,23 +38,24 @@
                                         </Style>
                                     </StackPanel.Resources>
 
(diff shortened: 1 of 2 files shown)


klaki892:
Those clear buttons are... Huge.
We okay with leaving them like that? High chance of accidental clicking?

bridCquinn:
I agree with @klaki892 . I think if you just make the buttons smaller and center them, that would probably be less likely of accidental clicking.

AlekStoyanov:
image
Is this better ?

PR #258, opened 2020-03-25
Position blocks horizontal

### LINKED ISSUE #242: Event alignment when clicking on a node
The farther right you scroll on the timeline, the more off center the event information pop-up is.

The events should show up right above the node that is clicked.

Example of bug:
image

bridCquinn:
This fixes the horizontal alignment of the blocks with the nodes. 
horizontal alignment
This shows the last node on the timeline.
Resolves #242

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml
+++ b/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml
@@ -153,7 +153,7 @@
                                         <RowDefinition Height="Auto"/>
                                         <RowDefinition Height="*"/>
                                     </Grid.RowDefinitions>
-                                    <StackPanel x:Name="Blocks" Grid.Row="0" Orientation="Horizontal" VerticalAlignment="Bottom"/>
+                                    <StackPanel x:Name="Blocks" Grid.Row="0" Orientation="Horizontal" VerticalAlignment="Bottom" Margin="0,0,200,0"/>
                                     <StackPanel x:Name="Timelines" Grid.Row="1" Orientation="Horizontal"/>
                                     <StackPanel x:Name="TimeStamps" Grid.Row="1" Orientation="Horizontal" Margin="0,97,0,0"/>
                                     <StackPanel x:Name="TicksBar" Grid.Row="1" Orientation="Horizontal" Margin="0,43,0,0"/>
--- a/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml
+++ b/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml
@@ -153,7 +153,7 @@
                                         <RowDefinition Height="Auto"/>
                                         <RowDefinition Height="*"/>
                                     </Grid.RowDefinitions>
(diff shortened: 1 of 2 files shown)


WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml
```diff
@@ -153,7 +153,7 @@
                                         <RowDefinition Height="Auto"/>
                                         <RowDefinition Height="*"/>
                                     </Grid.RowDefinitions>
-                                    <StackPanel x:Name="Blocks" Grid.Row="0" Orientation="Horizontal" VerticalAlignment="Bottom"/>
+                                    <StackPanel x:Name="Blocks" Grid.Row="0" Orientation="Horizontal" VerticalAlignment="Bottom" Margin="0,0,200,0"/>
```
klaki892: Could line be causing this?:
image

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -315,6 +315,26 @@ private TimelinePanel MakeTimelinePanel(DateTime beginDate, DateTime endDate)
...
+        /// Creates a TimelinePanel for Blocks
+        /// </summary>
+        /// <param name="beginDate">begin date of timeline</param>
+        /// <param name="endDate">end date of timeline</param>
+        /// <returns>TimelinePanel that can space Blocks according to time</returns>
+        private TimelinePanel MakeBlockPanel(DateTime beginDate, DateTime endDate)
+        {
+            TimelinePanel timelinePanel = new TimelinePanel
+            {
+                UnitTimeSpan = new TimeSpan(0, 0, 0, 10),
```
AlekStoyanov: Did you calculate this with proportions? Clever.

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml
AlekStoyanov: I think that the last block timeline extends past the last node timeline, thus this gap happens. To clarify, each block is 200px and each node is 20px, a timeline for the blocks is simply longer in order to fit the longer blocks. The blocks do overlap if the nodes are 1 second apart, but they still overextend past the node timeline they are associated with.

bridCquinn: So while working on this. I did realize that the last blocks were being cut off when it went past the node timeline. Since this is a direct hindrance to the information on the timeline, I thought it would be better to set a right margin so that the boxes weren't cut off. Alex got it right. 
In reference the the overlap, this issue was never supposed to fix the overlap between timelines, from what I understand. We decided this was a non issue at the moment, given the user can simply close one of the overlapping ones.

WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
bridCquinn: Given the nodes are 20px, spaced by 1, and the blocks were 200px, so increasing to 10 seemed to match the step. Honestly, there was a lot of trial and error to kind of understand how it all worked.

PR #260, opened 2020-03-26
Access parent filter with right click on an even't block

### LINKED ISSUE #256: Make Event Parent filter usable via a context-menu
Currently the `Event Parent` filter is unusable due to needing the ability for a user to specify the parent of an event. Without the user's ability to select an Event's parent the filter remains unusable other than internal testing. 

The only way for a user to start to use the parent of an event is to have the ability, from an event, to select the filter or an option to apply "Filter on Event's Parent" (do not use this verbatim wording)

It has been suggested that a context menu be enabled when clicking on nodes to provide the functionality to the user. Normally options such as these appear as a right click menu (and wont conflict with the functionality of #246) 

Enable the user to be able to filter by a Event's parent by an right click context menu.

Acceptance Criteria
- When the user right clicks on an expanded node's information, provide a context menu
- Upon clicking on an option to filter by event parent in a context menu, the timeline filters out all information except the information that came from the one selected ShellItem (the "events parent")
- The event filter on the GUI shows some identifier that the timeline is being filtered by an event parent

AlekStoyanov:
Adds the ability to use the filter that filters events by an even't parent.
The filter is used through a right click context menu on an event's block and gives the option to filter the timeline by this event's parent.
After the filter is applied it is indicated by stating the name of the parent that the timeline was filtered by in the Event Parent Filter filter box on the left hand side of the UI.
The filter can then be cleared by pressing the clear button.

image

image

Resolves #256

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
+++ b/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
@@ -28,10 +28,13 @@ namespace SeeShells.UI.Pages
 
         private TimeSpan maxRealTimeSpan = new TimeSpan(0, 0, 1, 0); // Max time in one timeline (1 min).
 
+        private static TimelinePage timelinePage;
+
         public TimelinePage()
         {
             InitializeComponent();
             BuildTimeline();
+            timelinePage = this;
         }
 
         private void AllStringFilter_TextChanged(object sender, TextChangedEventArgs e)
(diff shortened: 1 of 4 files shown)


WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -28,10 +28,13 @@ public partial class TimelinePage : Page
 
         private TimeSpan maxRealTimeSpan = new TimeSpan(0, 0, 1, 0); // Max time in one timeline (1 min).
 
+        private static TimelinePage timelinePage;
+
         public TimelinePage()
         {
             InitializeComponent();
             BuildTimeline();
+            timelinePage = this;
```
klaki892: Pretty sure this is a hard anti-pattern, but knowing the rest of the program i dont believe this will cause issues. 
Why was this static reference needed?

AlekStoyanov: The reason is that the `EventParentContextMenu_Click` method must be static, therefore I don't think there is another workaround to be able to update the timeline when that filter is applied since updating the timeline must be non-static.

PR #282, opened 2020-03-29
Possibe alignment modification

### LINKED ISSUE #244: Align information boxes vertically going bottom-up
Right now, the information boxes appear at the top of the scrollbar instead of the bottom. This can be an issue when you open a node with a lot of events and then click on a node with one or two events. The smaller node appears at first to have nothing pop up, but then you realize you have to scroll up to see anything.

# Normal view:
image

# Example of what goes wrong:
Open up a node with a lot of events:
image

You have to scroll up to see the 3 events for this node now:
image

# Possible solution:
- Keep track of what the largest number of events in a node are in the timelines.
- When clicking on a node, insert hidden/blank information boxes before inserting the actual information boxes. This would make them appear at the bottom .
-- Ex: The largest node holds 13 events. If a node with 2 events is clicked on, 11 hidden/blank information boxes are inserted before the actual information boxes about the 2 events are inserted. This aligns all the information boxes starting from the bottom instead of the top and prevents the user from having to scroll to see them.

szczura:
I had an idea for resolving #244 but it failed; however, I made this modification while working through it.

Instead of making a bunch of hidden boxes, we can just make one per stack and make it the height of the needed boxes combined. This helps with the massive memory issue.

To avoid the awkward stacking issue when two nodes are close to each other, we can hide the stacks that conflict with the ones we want to open most recently (see the pictures). I think this works better for clicking around and viewing a bunch of events. This also solves the "jittering" problem Aleks and Bridget were talking about (I only understood what that issue was when working on this today).

With this PR, the events should always be visible when you click on them.

# Pictures
Click on the 6 node:
image

Clicking on the 3 node:
image

As you can see, it hides the 6-node descriptions. When we click on the 6 node again, the 3 nodes hide and we can see the 6 ones.

If we don't like this idea, we can just close this PR without merging it.
If we like this idea, I think we can say it resolves #244 
(Parts of this code is also @AlekStoyanov 's; I worked off of what he started to make)

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
+++ b/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
@@ -193,7 +193,6 @@ namespace SeeShells.UI.Pages
                 }
 
                 List<Node.Node> nodeList = new List<Node.Node>();
-                List<TextBlock> blockList = new List<TextBlock>();
                 DateTime eventTime = App.nodeCollection.nodeList[0].aEvent.EventTime;
                 maxStackedNodes = 0;
                 int currentMaxStackedNodes = 0;
@@ -206,20 +205,16 @@ namespace SeeShells.UI.Pages
                     else
                     {
                         if(currentMaxStackedNodes > maxStackedNodes)
-                        {
                             maxStackedNodes = currentMaxStackedNodes;
-                        }
+
                         currentMaxStackedNodes = 1;
 
                         eventTime = node.aEvent.EventTime;
                     }
 
                     node.Style = (Style)Resources["Node"];
-                    if (node.Visibility == System.Windows.Visibility.Visible)
-                    {
+                    if (node.Visibility == Visibility.Visible)
                         nodeList.Add(node);
-                        blockList.Add(node.block);
-                    }
                     node.block.Visibility = Visibility.Collapsed;
 
                 }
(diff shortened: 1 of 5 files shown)


WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -583,16 +644,72 @@ private void DownloadClick(object sender, RoutedEventArgs e)
...
             else if(sender.GetType() == typeof(StackedNodes))
             {
+                unToggleEventsThatAreTooClose(((StackedNodes)sender).GetBlockTime());
                 ((StackedNodes)sender).ToggleBlock();
+
+                if (toggledNodes.Contains(sender))
+                    toggledNodes.Remove((StackedNodes)sender);
+                else
+                    toggledNodes.Add((StackedNodes)sender);
+            }
```
szczura: I wanna do a refactor eventually where StackedNodes and Node inherit from a class that inherits ToggleButton so that there's less of this type checking going on.

klaki892:
Are you seeing this when running it against the NTUser.dat in our Test Resources?
21f4c6fd-91e2-462a-9c03-6851df0d451f

szczura:
> 
> 
> Are you seeing this when running it against the NTUser.dat in our Test Resources?

Dang, I'll look into it. I was testing against NTUser & then switched to my regular shellbags when finishing. I must have changed something that crashes it.

szczura:
> 
> 
> > Are you seeing this when running it against the NTUser.dat in our Test Resources?
> 
> Dang, I'll look into it. 

Found the problem! If the stackedNode was the one that contained the most items, the alignment block was never created for it (since it didn't need it) so it was crashing on trying to hide/collapse it.

Issue #273, opened 2020-03-28
Create Margin in Blocks to stop cutting off part of the Info Block


bridCquinn:
Add a Margin the size of the hovered over block to the Blocks Stackpanel.

Issue #257, opened 2020-03-24
Streamline Parsing configurations into an "Advanced" section on Home page


klaki892:
When running and using SeeShells right now, you have to go through 3 pop up boxes before you're allowed to parse. (The confirmations introduced in #197)  
Seeing configuration file options upon first startup could be overwhelming to a brand new user who only:
1. just downloaded the program
2. wants to immediate see how it operates

To support the simplest path of execution for the program  and to follow the principle of [Making the Common Case Fast](https://www.elsevier.com/connect/8-great-ideas-in-computer-architecture), all options supporting configurations should be moved/hidden so they no longer need to be interacted with unless the user explicitly wants to do so. 

As a baseline these requirements include:
- By default don't show the user configuration options for the 3 configuration files (the filepaths, browsing, updating each file)  (e.g. a checkbox that says "Advanced", "Advanced Configurations", "Advanced Options", etc.)
- Don't show the configuration warnings unless the user has shown interest in configuring how the application runs. (they've "unhidden" whatever method is hiding the advanced configurations)

Acceptance Criteria:
- Optional configurations for parsing are hidden/collapsed until a user makes an explicit choice to see such options. 
- Warnings about optional configurations (and possible misconfigurations) are only shown if the optional configurations are also shown

klaki892:
Reopening as acceptance Criteria wasn't met previously.
 As of the current master, Configuration warnings still appear even if the user cant see the contents of the advanced section

Issue #188, opened 2020-02-26
Rescaling of WPF UI 


klaki892:
Currently the does not scale properly due to various constraits and behaviors from the existing UI. Because of these oddities in scaling, UI elements might appear to big or wont scale to certain resolutions that were not originally designed against. Examples:
![image](https://user-images.githubusercontent.com/17878901/75397805-ca0ef180-58c5-11ea-8cc8-3ebd861c3680.png)

![image](https://user-images.githubusercontent.com/17878901/75397812-d2672c80-58c5-11ea-95ae-346ecf5428c9.png)

Acceptance Criteria:
- Windows and elements need to be scaled such that the entire window is taken up regardless of resolution
- elements of UI need to have min/max scaling such that they are still readable at _reasonable_ resolutions and aspect ratios.

bridCquinn:
Remastering the Graphical WPF:
- Take out the tabs
- Resize the GUI for smaller screen
- Change colors to be easily readable

klaki892:
> * Take out the tabs

This task duplicates one of the acceptance criteria of #189

PR #286, opened 2020-03-29
Remaster gui

bridCquinn:
Changed the home Page:
 Added the advanced settings area
 Moved elements around 
The text colors were changed for ease of reading
Added Margin for Blocks


szczura:
This does not accomplish the acceptance criteria for the task assigned:

- Windows and elements need to be scaled such that the entire window is taken up regardless of resolution
- elements of UI need to have min/max scaling such that they are still readable at reasonable resolutions and aspect ratios.

image
image

szczura:
Can you add screenshots when you make updates?

WPF/SeeShells/SeeShells/UI/Node/InfoBlock.cs
```diff
@@ -67,7 +67,7 @@ public void ToggleInfo()
             }
             else
             {
-                this.Width = 450;
+                this.Width = Double.NaN;
```
szczura: This isn't going to work as the solution to the info-boxes being cut off. 

A user can't see all the information:
image

bridCquinn:
Beginning Page
This is when it first opens

everything open
This is with everything open

everything open
This is the vertical view

klaki892:
> Beginning Page
> This is when it first opens

I like this design, but all of this empty space with the word "Offline" Now just seems...wrong? A random person wouldnt know what we meant by just the word "offline"
 
**@Team** wouldn't this make more sense if we made it a RadioButton with the options "Live Registry" and "Offline Registry" or "Registry from file(s)"?

bridCquinn:
This is the rescaled GUI timeline page.
timlinepage

This is now how the hover is done.
hover fix

What was done:
- The timeline and home page now adapts to the size of the screen which the user may put it in. 
- The size of the window which opens has been modified to give the optimal viewing experience
- The min sizes have been set to eliminate a UI too small to read
- The Home page now has an advanced section as to not overwhelm the new user with configuration
- The hovering has been changed to have a wrapping functionality

The Problem (Has now been overcome):
The issue with the UI never taking up the entire space of the screen was caused by the tool bar. It was wrapping the entire window in a WrapPanel, so I changed it to a dock panel and simply docked it to the top. After this was done, all of the grids gained their normal functionality.

bridCquinn:
radio button reg
This is what the opening page looks like now with the radio buttons (the online registry is the pre-selected choice, to stay with the idea of only having to press the Parse button for an online registry)

radiobutton open
This is what it looks like when everything is open.

@klaki892 Thanks for the suggestion!

bridCquinn:
Resolves #273 . Resolves #257 . Resolves #188 . Resolves #290

WPF/SeeShells/SeeShells/UI/Node/InfoBlock.cs
bridCquinn: This has now been changed in the updates below, pictures are provided.

klaki892:
> Resolves #273 . Resolves #257 . Resolves #188 . Resolves #290

The Resolves keyword only works in the opening message of a PR
You can add these issues with the sidebar saying "linked issues"

Issue #259, opened 2020-03-25
Lack of Time Zones in events lead to misleading Timelines


klaki892:
![image](https://user-images.githubusercontent.com/17878901/77571795-5223ef00-6ea4-11ea-99c2-ad9e9c28c097.png)
(Notice that it appears these events happened "in the future" (4 hours ahead) )

Currently all events and their `DateTime` objects are processed in UTC, as represented in the backend by the registry items.  However, SeeShells currently does not convert this Timezone to the User's timezone running the program, nor does it inform the user that UTC is the timezone being displayed in the events. 

This can cause seriously misleading data, such as "future events" or prevent the correlation of events with other forensics recovery tools.  The program should either:
- Notify in the events (and the export of such events)  what the timezone of the action is (preferred)
- Convert to all displayed Time Zones to the currently running system. 

Steps to reproduce:
1. Note the current time
2. Perform an action that creates a Shellbag (i.e. creating a folder)
3.  Run SeeShells and perform an Online parse.
4. observe the events appear to happen in the future or the past depending on your timezone (if EST, the event will appear 4 hours later)

Acceptance Criteria:
Either:
- DateTimes in events display the timezone both in the program and in export formats
OR
- All DateTimes are converted to the currently running system's Timezone when shown in the program and exports show the timezone (e.g. HTML output)

PR #287, opened 2020-03-29
Timezone addition

yara-58:
My solution to this bug was to convert all the datetimes from UTC to the Users local timezone on their system before they are added as part of an event. This was made possible by Klayton making all the registry DateTimes UTC, whereas before they were based on the user's computer while shellbags were based on Utc. I also added the Local Time Zone on the computer to the HTML output. I tested it using multiple timezones on my computer and the EventTimes change to that of the current timezone. 

proofofoperation

current html output

this closes issue #259

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/EventParser.cs
+++ b/WPF/SeeShells/SeeShells/UI/EventParser.cs
@@ -17,6 +17,7 @@ namespace SeeShells.UI
         public static List<IEvent> GetEvents(List<IShellItem> shells)
         {
             List<IEvent> eventList = new List<IEvent>();
+            TimeZoneInfo time = TimeZoneInfo.Local;
             foreach (IShellItem item in shells)
             {
                 IDictionary<String, String> parser = item.GetAllProperties();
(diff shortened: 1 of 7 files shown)


yara-58:
actual HTML output

sorry the last comment had the wrong HTML output image. This is the correct one.

yara-58:
proofofhtmloutput

this is the new html output with the correct times in the details

klaki892:
> proofofhtmloutput
> 
> this is the new html output with the correct times in the details

Are the dates also consistent in the expandable boxes of the Timeline?

yara-58:
window displaying correct timezone information

They are now

WPF/SeeShells/SeeShells/UI/EventParser.cs
```diff
@@ -17,6 +17,7 @@ public static class EventParser
         public static List<IEvent> GetEvents(List<IShellItem> shells)
         {
             List<IEvent> eventList = new List<IEvent>();
+            TimeZoneInfo time = TimeZoneInfo.Local;
```
szczura: Why do we use TimeZoneInfo.Local here while we get TimeZone.CurrentTimeZone in Event.cs? Does this affect anything? (Commenting this here, but it shows up a couple of times; just wondering what the difference is between them)

yara-58: this variable is used because In order to convert from UTC i have to tell it what the datetime is converting to which is the local timezone but that has to be of type TimeZoneInfo for the conversion function. The TimeZone.CurrentTimeZone if of type TimeZone and is used in the html output to display the timezone on a user's computer.

PR #288, opened 2020-03-29
Prevents program crash when trying to write to a CSV file locked for editing

### LINKED ISSUE #279: SeeShells Crashes when writing to a CSV file which is locked for editing
It has been observed that SeeShells will crash if it attempts to write to a file which is already locked for editing by another process (e.g. Excel when it has the same file open)

Typical behavior for a program in such a situation  situation is to warn the user the file can't be saved (and potentially the reason why) 

Steps to Reproduce:
1. Perform any type of Parse operation
2. Save the output as a CSV file
3. Open the CSV file via another program or process which will write-lock the file.
4. Attempt to output the CSV data again in SeeShells
5. Observe SeeShells crashes due to not catching the exception produced from trying to open with write access to a write-locked file.

Acceptance Criteria:
- SeeShells no longer crashes when attempting to save to a write-locked CSV file.
- The User is informed that the save to CSV operation did not complete successfully

AlekStoyanov:
If the user tries to export CSV output to an already existing CSV file that is currently opened by another process for editing, the program will notify the user with a message box, log the exception and proceed without exporting the CSV file.

image

Resolves #279

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
+++ b/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
@@ -653,7 +653,8 @@ namespace SeeShells.UI.Pages
                         }
                         catch (IOException ex)
                         {
-                            System.Windows.MessageBox.Show("The file: \"" + name2 + "\" is being used by another process.");
+                            System.Windows.MessageBox.Show("The file: \"" + name2 + "\" is being used by another process.\n" +
+                                "Select a different file name or close the process to save the CSV.", "Cannot save file", MessageBoxButton.OK, MessageBoxImage.Exclamation);
                             logger.Info(ex, "The file: \"" + name2 + "\" is being used by another process.\n" + ex.ToString());
                             return;
                         }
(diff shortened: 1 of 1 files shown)


WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
```diff
@@ -639,7 +640,26 @@ private void DownloadClick(object sender, RoutedEventArgs e)
...
+                        {
+                            using (FileStream stream = file.Open(FileMode.Open, FileAccess.Read, FileShare.None))
+                            {
+                                stream.Close();
+                            }
+                        }
+                        catch (IOException ex)
+                        {
+                            System.Windows.MessageBox.Show("The file: \"" + name2 + "\" is being used by another process.");
+                            logger.Info(ex, "The file: \"" + name2 + "\" is being used by another process.\n" + ex.ToString());
```
szczura: Could we add on to this message: "Select a different file name or close the process to save the CSV." Just to make it clear that BECAUSE the file is being used by another process, we were not able to save the CSV.

szczura: Could we add on to this message: "Select a different file name or close the process to save the CSV." Just to make it clear that BECAUSE the file is being used by another process, we were not able to save the CSV.

szczura: Oh yeah, and can you also add a caption and messageboximage.

AlekStoyanov: image

PR #289, opened 2020-03-30
Warn when running without proper permissions

### LINKED ISSUE #237: SeeShells crashes on Online parsing if the user isnt an administrator
Because of the need to look into registry locations above the Users level for online parsing, a requirement of beginning an Online parse requires that the program is running with administrator privileges.

Currently, the program does not prompt or warn of this restriction, and the program will crash after hitting the parse button with no indication to the user what went wrong.

Steps to Reproduce:
1. Run SeeShells.exe **without** administrator privileges (e.g. non-admin user, not accepting the UAC dialog) 
2. Click the Parse button
3. Observe the program crashes

Two options that exist for handling this bug:
1. Check and inform the user that they must be running as an administrator to continue with an online parse operation, prevent the operation if they are not.
2. When the user attempts to parse and doesn't have privileges, invoke the UAC dialog to escalate the program's privileges to continue to the operation. 

Acceptance Criteria:
- SeeShells gracefully handles/tells the user that administrative privileges are required for online parsing
- Seeshells **does not** crash if online parsing is attempted without admin rights.

szczura:
Resolves #237 
image


WPF/SeeShells/SeeShells/UI/Pages/Home.xaml.cs
```diff
@@ -285,10 +285,23 @@ private async void ParseButton_Click(object sender, RoutedEventArgs e)
...
+            try
+            {
+                App.ShellItems = await ParseShellBags();
+                List<IEvent> events = EventParser.GetEvents(App.ShellItems);
+                App.nodeCollection.ClearAllFilters();
+                App.nodeCollection.nodeList.AddRange(NodeParser.GetNodes(events));
+            }
+            catch (System.Security.SecurityException)
+            {
+                MessageBox.Show("Please exit the program and run SeeShells as an administrator to parse a live registry.", "Access Denied", MessageBoxButton.OK, MessageBoxImage.Error);
```
klaki892: We have `showErrorMessage()` so we dont manually invoke messageboxes

klaki892:
Not that it has to be done, but isn't it possible to invoke UAC from inside the application so it could continue as normal?

szczura:
> 
> 
> Not that it has to be done, (because this solution does everything we need it to) but isn't it possible to invoke UAC from inside the application so it could continue as normal with elevated privileges?

Didn't realize this was a thing WPF could do; we can request that Seeshells can only be run with administrative privileges by adding & modifying the app manifest file. We can't technically invoke UAC within the application & continue because that would launch a new thread and restart the program.

klaki892:
> > Not that it has to be done, (because this solution does everything we need it to) but isn't it possible to invoke UAC from inside the application so it could continue as normal with elevated privileges?
> 
> Didn't realize this was a thing WPF could do; we can request that Seeshells can only be run with administrative privileges by adding & modifying the app manifest file. We can't technically invoke UAC within the application & continue because that would launch a new thread and restart the program.

Requesting that the entire application be run as an administrator (and not letting the user use if they arent) is too heavy handed. 
We dont need admin privileges for the entire application, we only need it for the Online parsing process. Offline parsing needs no such privileges and we just stopped the user from being able to use that functionality. 
Consider if someone was using this application (from their flashdrive) on their work computer. normal users dont usually have admin access to their box and now the tool is unusable entirely.

If we cant do it within the app we might as well just leave it at the error message.

szczura:
> 
> 
> > > Not that it has to be done, (because this solution does everything we need it to) but isn't it possible to invoke UAC from inside the application so it could continue as normal with elevated privileges?
> > 
> > 
> > Didn't realize this was a thing WPF could do; we can request that Seeshells can only be run with administrative privileges by adding & modifying the app manifest file. We can't technically invoke UAC within the application & continue because that would launch a new thread and restart the program.
> 
> Requesting that the entire application be run as an administrator (and not letting the user use it if they arent) is too heavy handed.
> We dont need admin privileges for the entire application, we only need it for the Online parsing process. Offline parsing needs no such privileges and we just stopped the user from being able to use that functionality.
> Consider if someone was using this application (from their flashdrive) on their work computer. normal users dont usually have admin access to their box and now the tool is unusable entirely. (and they probably werent about to analyze their own work computer)
> 
> If we cant do it within the app we might as well just leave it at the error message.

That makes sense. I'll remove the manifest & keep it at the error message.

PR #298, opened 2020-03-31
Bug Fix

bridCquinn:
Info Blocks were being cut off and now there is a margin. I added a 350 margin to the right. This will work for hovered blocks


klaki892:
image

Issue #281, opened 2020-03-28
Obtain time based registry Metadata for shellitems during offline parsing


klaki892:
This is a SubTask of #276

Work done from #271 has enabled this task to be independent due to the refactoring done in that PR. 

Currently Offline Hive Parsing of the registry does not pull the necessary information out of registry hives to be compatible with the output expected.

Offline Hive parsing should be able to obtain the same pieces of information which have come to be expected during an live/Online hive parsing. All the relevant information is gathered in the `RegistryKeyWrapper`

Acceptance Criteria:
- Offline Hive parsing populates `RegistryKeyWrapper` with enough information to be currently outputted by the `RegistryShellItemDecorator` 

_Originally posted by @klaki892 in https://github.com/RickLeinecker/SeeShells/issues/276#issuecomment-605517950_

PR #300, opened 2020-04-01
Modifies the offline parser to retrieve additional time based registry metadata for shellitems

AlekStoyanov:
The new modifications to the offline parser make it build registry key wrappers while parsing and enable it to gather time based registry metadata for shellitems such as:
- SlotModifiedDate
- LastRegistryWriteDate
- ShellbagPath
- Parent

Resolves: #281

### DIFF
--- a/WPF/SeeShells/SeeShells/ShellParser/Registry/RegistryKeyWrapper.cs
+++ b/WPF/SeeShells/SeeShells/ShellParser/Registry/RegistryKeyWrapper.cs
@@ -142,7 +142,7 @@ namespace SeeShells.ShellParser.Registry
                     }
                 }
                 var shellbagKey = hive.GetKey(ShellbagPath);
-                SlotModifiedDate = shellbagKey.LastWriteTime.Value.DateTime;
+                SlotModifiedDate = shellbagKey.LastWriteTime.Value.LocalDateTime;
             }
             catch (Exception ex)
             {
@@ -150,8 +150,7 @@ namespace SeeShells.ShellParser.Registry
             }
 
             //obtain the date the registry last wrote this key
-            LastRegistryWriteDate = registryKey.LastWriteTime.Value.DateTime;
-
+            LastRegistryWriteDate = registryKey.LastWriteTime.Value.LocalDateTime;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


WPF/SeeShells/SeeShells/ShellParser/Registry/RegistryKeyWrapper.cs
```diff
@@ -108,5 +110,47 @@ private void AdaptWin32Key(Microsoft.Win32.RegistryKey registryKey)
...
+                }
+                SlotModifiedDate = registryKey.LastWriteTime.Value.DateTime;
+            }
+            catch (Exception ex)
+            {
+                logger.Trace(ex, $"NodeSlot was not found for registry key at {RegistryPath}");
+            }
+
+            //obtain the date the registry last wrote this key
+            LastRegistryWriteDate = registryKey.LastWriteTime.Value.DateTime;
```
klaki892, line 152: `SlotModifiedDate` and `LastRegistryWriteDate` do not both pull from the same key.

`SlotModifiedDate` pulls the LastWriteTime out of the registrykey at `ShellBagPath`. (see `AdaptWin32` and how it uses the shellbagpath , not the path of the current key we are adapting)

AlekStoyanov:
> Check to make sure that a parsed shellbag has the same data when run both offline and online.
> (i dont mean make a unit test, just a visual confirmation)

So it seems there are some inconsistencies between online and offline parsing.
Here is an example:
This from an offline parse
image
This is from a live parse
image

I am not sure why.

klaki892:
> > Check to make sure that a parsed shellbag has the same data when run both offline and online.
> > (i dont mean make a unit test, just a visual confirmation)
> 
> So it seems there are some inconsistencies between online and offline parsing.
> Here is an example:
> This from an offline parse
> (image)
> This is from a live parse
> (image)
> I am not sure why.

Other than the Inconsistent Last Registry Write Times, all other information which isn't matching one another should be moved to a new issue.

szczura:
> 
> 
> Resolve Offline parsing `LastRegistryWriteTime` not being consistent with the Online parsing time

I'm holding off on approving until this is updated.

Issue #161, opened 2020-02-19
Packaging WPF executable for Release


klaki892:
Currently the `SeeShells.exe` cannot execute independently in a folder by itself, it relies on multitudes of external .dll files when doing a debug or release build. As per the requirements of the sponsor there should be a way to hide or package the external dependencies inside the executable so that only the executable is needed to run the program.

Acceptance Criteria:
-the WPF application can run in any location on a supported windows machine without external dll's nearby
-runs on flashdrive

klaki892:
[A stackoverflow question answer on this topic ](https://stackoverflow.com/a/20306095) shows a possible open source package that could assist in this delivering process.

Issue #137 is also a part of this issue

PR #310, opened 2020-04-04
The SeeShells .exe now runs on its own

yara-58:
In order to accomplish this I just installed the Costra.Fody nuget package. This automatically embeds all references into the main assembly. After installing the package I rebuilt the solution and copied just the .exe onto my desktop,  ran it from there, and it works properly without needing any external .dll files.

I also coped the executable into a USB and ran the program from the USB on two other computers and It ran as expected. 

The program does still need to be run as an administrator in order to parse the live hive which is as its always been just put this here as a reminder if you're testing this on your computers.


klaki892:
please dont forget to add reviewers and link the issue your resolving.

I've noticed that the log file is no longer created, so we are no longer getting logs of the actions on SeeShells. It'll be hard for anyone in the future to report errors unless its from a debugging session.

yara-58:
resolves issue #161

I looked into the log files not being generated and I cannot find a way around that and still be able to run the executable without external dlls. This could possibly be made into a new issue if need be. I am trying to look for another solution and if you guys find anything please let me know.

klaki892:
> resolves issue #161
> 
> I looked into the log files not being generated and I cannot find a way around that and still be able to run the executable without external dlls. This could possibly be made into a new issue if need be. I am trying to look for another solution and if you guys find anything please let me know.

- [Talk on Fody.Costura about this issue.](https://github.com/Fody/Costura/issues/155)
- [Second Issue Mentioning the issue (and how to fix)](https://github.com/NLog/NLog/issues/165)
- [The fix.](https://github.com/NLog/NLog/wiki/Configure-from-code)

yara-58:
A Log file is now generated when the user runs the executable in the same directory where the executable is stored.

WPF/SeeShells/SeeShells/App.xaml.cs
```diff
@@ -64,6 +65,19 @@ private void Toolbar_Loaded(object sender, RoutedEventArgs e)
...
+
+            var config = new NLog.Config.LoggingConfiguration();
+
+            // Targets where to log to: File and Console
+            var logfile = new NLog.Targets.FileTarget("logfile") { FileName = "${basedir}/log.txt" };
+            var logconsole = new NLog.Targets.ConsoleTarget("logconsole");
+
+            // Rules for mapping loggers to targets            
+            config.AddRule(LogLevel.Info, LogLevel.Fatal, logconsole);
+            config.AddRule(LogLevel.Debug, LogLevel.Fatal, logfile);
```
klaki892, line 77: These arent the lower end log levels we use. See the Nlog.Config file.
Although changing them is a subjective decision for the team. I'd err on giving as much info as possible if we arent providing the means to reload the config from a file

PR #316, opened 2020-04-06
Navigation/refresh doesn't display 404 page.

### LINKED ISSUE #236: Website/Github-Pages cant handle url navigation / webpage reloads
Currently when you navigate to a specific page other than the home page, or attempt to refresh a page while not on a home page, Github pages returns a 404 because our website is a single-page application and github-pages by default will attempt to navigate to the nonexistent directory created by our pathing.

Steps to Reproduce:
1. [Go to the site](https://rickleinecker.github.io/SeeShells)
2. navigate to any subdirectory/page (e.g. [login page](https://rickleinecker.github.io/SeeShells/login))
3. attempt to refresh the page
4. notice a 404 error

Acceptance Criteria:
- When navigating to  or reloading pages other than the home page, the proper page should be displayed, not a 404 page.

szczura:
Resolves #236

This was working properly on my forked repo. Refreshing on About kept me on the about page. Typing in blahblah/SeeShells/team took me to the team page instead of 404.

Solution came from here:
https://stackoverflow.com/questions/47677220/vuejs-history-mode-with-github-gitlab-pages/50259501#50259501


szczura:
> 
> 
> What happens if someone tries to directly navigate to a administrative section while not logged in?

I didn't think about that; good point. They'll be able to see the page, but anything they try to do won't work since the administrative API endpoints check if they're logged in before doing anything. Should I modify this to not allow people to see the pages if they're not logged in?

klaki892:
> > What happens if someone tries to directly navigate to a administrative section while not logged in?
> 
> I didn't think about that; good point. They'll be able to see the page, but anything they try to do won't work since the administrative API endpoints check if they're logged in before doing anything. Should I modify this to not allow people to see the pages if they're not logged in?

_should _we?__ Yes
_do we have to?_ No

The real use case here is if someone reloads one of the administrative pages after an extended period of time (from a long open tab, or they just went to their history and clicked on a specific one) and their cookie expired.  They'd see the page and get a bunch of errors on attempting to do anything.
Ideally we should just redirect to the login page if such a situation occurred. If not, we're betting on the user eventually figuring out they have to re-log after receiving errors

szczura:
> 
> 
> > > What happens if someone tries to directly navigate to a administrative section while not logged in?
> > 
> > 
> > I didn't think about that; good point. They'll be able to see the page, but anything they try to do won't work since the administrative API endpoints check if they're logged in before doing anything. Should I modify this to not allow people to see the pages if they're not logged in?
> The real use case here is if someone reloads one of the administrative pages after an extended period of time (from a long open tab, or they just went to their history and clicked on a specific one) and their cookie expired. They'd see the page and get a bunch of errors on attempting to do anything.
> Ideally we should just redirect to the login page if such a situation occurred. If not, we're betting on the user eventually figuring out they have to re-log after receiving errors

Oh, it does redirect to the login if they try to do something and aren't logged in.

klaki892:
> > The real use case here is if someone reloads one of the administrative pages after an extended period of time (from a long open tab, or they just went to their history and clicked on a specific one) and their cookie expired. They'd see the page and get a bunch of errors on attempting to do anything.
> > Ideally we should just redirect to the login page if such a situation occurred. If not, we're betting on the user eventually figuring out they have to re-log after receiving errors
> 
> Oh, it does redirect to the login if they try to do something and aren't logged in. It just wouldn't redirect until they tried to do something on an admin page.

Would this redirect on attempting to post information or get information? 
This might be fine, but if it only redirects on posts there's a chance someone entered information and they'd just lose it on redirect.

szczura:
> 
> 
> > > The real use case here is if someone reloads one of the administrative pages after an extended period of time (from a long open tab, or they just went to their history and clicked on a specific one) and their cookie expired. They'd see the page and get a bunch of errors on attempting to do anything.
> > > Ideally we should just redirect to the login page if such a situation occurred. If not, we're betting on the user eventually figuring out they have to re-log after receiving errors
> > 
> > 
> > Oh, it does redirect to the login if they try to do something and aren't logged in. It just wouldn't redirect until they tried to do something on an admin page.
> 
> Would this redirect on attempting to post information or get information?
> This might be fine, but if it only redirects on posts there's a chance someone entered information and they'd just lose it on redirect.

Only on post. I figured it's fine to get information when not logged in, but it would make sense to check on get info to redirect before they try to input something and lose data. I'll make a new issue for it.

PR #323, opened 2020-04-07
Check if user is authorized before loading admin pages

### LINKED ISSUE #318: Redirect user to login if accessing admin page & session is expired or user is not logged in
Checking if the user is logged in on GET requests used in the admin side of the website would allow us to redirect the user to the login page if their session expires earlier. There would be no risk of losing data from an attempted POST request (and then redirect) since the expired session would be noticed right when they access a page before they try to input anything.

Acceptance criteria:
- GET requests used in the admin side of the website check if the user is logged in/if session is expired
- If expired, users are redirected to the login page.

szczura:
Resolves #318 

This is a different method than the one described in the issue, but it solves the problem we were trying to solve in a better way imo. Instead of adding checks to multiple GET requests (which would mess with the WPF trying to pull information), we can just check if the session is valid before loading any admin pages.

Going to an admin page without being logged in or with an expired session redirects you to the login page.


website/src/components/ApproveUserPage.vue
```diff
@@ -14,6 +14,28 @@
...
+
+                if (result.success != 1) {
+                        this.$session.destroy();
+                        this.$router.push('/SeeShells/login');
+                        location.reload();
+                }
+            }
+            catch (err) {
+                console.info(err);
+            }
```
klaki892, line 37: This appears to have just been copy pasted into different files.
This can just be one utility function.

szczura: So googling utility function led me to "mixins" - apparently the way Vue handles this. I'll modify it to fit that.

PR #324, opened 2020-04-07
Adds a context menu on right click

### LINKED ISSUE #303: Right clicking on an event opens the right context menu and popped out window
Currently the controls for a Node consist of two things:
- clicking opens the pop-out window 
- Right clicking opens a context menu providing the option to use the Event Parent Filter

Right now right clicking performs both of the actions above while left clicking only creates the pop-out window. The pop-out window functionality while right clicking is not intended functionality and should be restricted to the left click action only. 

Steps to Reproduce:
1. Get to the timeline page which has at least 1 event displayed on the timeline.
2. Click on a Node
3. Right click on the event summary to see it performs two actions:
   - Opens the pop out window 
   - Display's the context menu

Acceptance Criteria:
- Right clicking an event no longer opens the pop out menu

### LINKED ISSUE #304: Improve User Experience For Clicking Options on Event Infoboxes
The bug presented in #303 has opened the discussion to avenues that could improve the user experience when clicking on event infoboxes / expanded info boxes.

Currently the intended functionality is:
- Left clicking opens the pop-out window
- Right clicking opens a context menu providing the option to use the Event Parent Filter

The following has been suggested as means of improving the User's experience for the provided clicking controls:
- Add another option in the right click context menu to"pop out" the information in the infobox.
- Doing the above, but also enabling the context menu for left clicks, so that only the context-menu occurs on both left and right clicking of the events

Additional suggestions may be added to the comments.

Acceptance Criteria:
- Discuss which avenue to take (if any) to improve the clicking functionality for events
   - Implement the chosen decision from the team.

bridCquinn:
This has the option to open a new window with event data or filter with that certain event. Resolves #303. Resolves #304 
popup

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Node/NodeParser.cs
+++ b/WPF/SeeShells/SeeShells/UI/Node/NodeParser.cs
@@ -57,8 +57,9 @@ namespace SeeShells.UI.Node
             block.Text += aEvent.EventType + "\n";
             block.MouseEnter += Pages.TimelinePage.HoverBlock;
             block.MouseLeave += Pages.TimelinePage.HoverBlock;
-            block.MouseDown += Pages.TimelinePage.ClickBlock;
-            block.MouseUp += Pages.TimelinePage.ClickBlock;
+            block.MouseRightButtonDown += Pages.TimelinePage.ClickBlock;
+            block.MouseRightButtonUp += Pages.TimelinePage.ClickBlock;
+            block.MouseLeftButtonDown += Pages.TimelinePage.LeftPopOut;
 
             ContextMenu contextMenu = new ContextMenu();
 
--- a/WPF/SeeShells/SeeShells/UI/Node/NodeParser.cs
+++ b/WPF/SeeShells/SeeShells/UI/Node/NodeParser.cs
@@ -58,13 +58,22 @@ namespace SeeShells.UI.Node
             block.MouseEnter += Pages.TimelinePage.HoverBlock;
             block.MouseLeave += Pages.TimelinePage.HoverBlock;
             block.MouseDown += Pages.TimelinePage.ClickBlock;
+            block.MouseUp += Pages.TimelinePage.ClickBlock;
 
             ContextMenu contextMenu = new ContextMenu();
+
             MenuItem parentFilter = new MenuItem();
             parentFilter.Header = "Filter for events with same parent";
             parentFilter.Click += Pages.TimelinePage.EventParentContextMenu_Click;
             parentFilter.Tag = aEvent.Parent;
             contextMenu.Items.Add(parentFilter);
+
+            MenuItem popOut = new MenuItem();
+            popOut.Header = "Create new window for event";
+            popOut.Click += Pages.TimelinePage.popOut;
+            popOut.Tag = block;
+            contextMenu.Items.Add(popOut);
+
             block.ContextMenu = contextMenu;
         }
     }
--- a/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
+++ b/WPF/SeeShells/SeeShells/UI/Pages/TimelinePage.xaml.cs
@@ -794,7 +794,7 @@ namespace SeeShells.UI.Pages
         }
 
         /// <summary>
-        /// This pops up a new window with information from the block.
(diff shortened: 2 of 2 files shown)


klaki892:
> This is not what the intention of #303 was. It technically hits the acceptance criteria but introduces a new issue.
> 
> Left clicking should still open the pop out window. Right clicking only should open the context menu.
> 
> If we leave it like this, the user will get very annoyed at having to click twice to open any windows.

Im not sure if this got lost in translation, but the acceptance criteria missing the bit about the left click was **not** an oversight. 
I questioned this behavior (context menu on both left and right clicks) while we discussed this bug/issue and the team decided to let the left and right click opens the context menu.

szczura:
> 
> Im not sure if this got lost in translation, but the acceptance criteria missing the bit about the left click was **not** an oversight.
> I questioned this behavior (context menu on both left and right clicks) while we discussed this bug/issue and the team decided to let the left and right click open the context menu.

I'm pretty sure we agreed on the opposite. It even says this in the issue: "Right now right clicking performs both of the actions above while left clicking only creates the pop-out window. The pop-out window functionality while right clicking is not intended functionality and should be restricted to the left click action only."

bridCquinn:
From our meetings and discussion, I thought we decided that the natural flow of things is to only use a right click. Given it is only two options, using a left click would just be confusing to the user.

szczura:
> 
> 
> From our meetings and discussion, I thought we decided that the natural flow of things is to only use a right click. Given it is only two options, using a left click would just be confusing to the user.

Both left and right click are opening the context menu.

bridCquinn:
Okay, I can fix that, but I would like a clear idea of what we all agreed, because I feel like all of us have different ideas. So what I understand is we only want right clicks to open the context window which holds both options and the left click does nothing. Please elaborate on what you want.

szczura:
> 
> 
> Okay, I can fix that, but I would like a clear idea of what we all agreed, because I feel like all of us have different ideas. So what I understand is we only want right clicks to open the context window which holds both options and the left click does nothing. Please elaborate on what you want.

My view is exactly what the issue says:

> Right now right clicking performs both of the actions above while left clicking only creates the pop-out window. The pop-out window functionality while right clicking is not intended functionality and should be restricted to the left click action only.
> 
> Acceptance Criteria:
> - Right clicking an event no longer opens the pop out menu

I don't understand why we would get rid of the left click. It sounds like we may need to have a call about this. I'm not sure where the miscommunication happened.

bridCquinn:
I understand the issue, but from what we discussed, I understood it differently. I will change it so that the pop out only happens on a left click and the context menu happens on the right.

klaki892:
> > Okay, I can fix that, but I would like a clear idea of what we all agreed, because I feel like all of us have different ideas. So what I understand is we only want right clicks to open the context window which holds both options and the left click does nothing. Please elaborate on what you want.
> 
> My view is exactly what the issue says:
> 
> > Right now right clicking performs both of the actions above while left clicking only creates the pop-out window. The pop-out window functionality while right clicking is not intended functionality and should be restricted to the left click action only.
> > Acceptance Criteria:
> > 
> > * Right clicking an event no longer opens the pop out menu
> 
> I don't understand why we would get rid of the left click. It sounds like we may need to have a call about this. I'm not sure where the miscommunication happened.

You're missing the suggestions as taken from the metting, but mentioned in #304 

> The following has been suggested as means of improving the User's experience for the provided clicking controls:
> 
> -  Add another option in the right click context menu to"pop out" the information in the infobox.
> - Doing the above, but also enabling the context menu for left clicks, so that only the context-menu occurs on both left and right clicking of the events
> Additional suggestions may be added to the comments.

szczura:
> 
> You're missing the suggestions as taken from the metting, but mentioned in #304
> 
> > The following has been suggested as means of improving the User's experience for the provided clicking controls:
> > 
> > * Add another option in the right click context menu to"pop out" the information in the infobox.
> > * Doing the above, but also enabling the context menu for left clicks, so that only the context-menu occurs on both left and right clicking of the events
> >   Additional suggestions may be added to the comments.

I remember us talking about it but not agreeing about it. I feel like if we agreed on it then we should have changed #303. @bridCquinn before you change anything I think we should do a quick call or vote with everyone.

PR #325, opened 2020-04-08
Adds single node navigation capability

### LINKED ISSUE #153: Button to go to next event
For some of the sliders, when the events are spaced apart a lot, it is difficult to find the events. A "next event" or "previous event" button would be helpful to hop to different parts of the timeline without missing any events.

AlekStoyanov:
Adds two UI buttons that allow a user to navigate forwards and backwards trough the nodes, one node at a time. (I did not use the arrow keys binding, since an initial click to focus is required and also the arrows are use by the timeline scrollbar by default.)

Navigation starts from the second node when the  forwards button is pressed. (The navigation is not circular. When the end is reached, pressing the button does noting. Same applies in the opposite direction)

When a node is brought into view, it also changes color to signify that it has been navigated to.
When a user left clicks on a node, the navigation makes that node the node to start navigating from. Otherwise navigation starts from the second node or wherever it was left off.

Tested and works when the timeline is rebuilt. (Therefore it works when it's filtered as well) 

Nav

Resolves #153

### DIFF
--- a/WPF/SeeShells/SeeShells/UI/Node/NodeNavigation.cs
+++ b/WPF/SeeShells/SeeShells/UI/Node/NodeNavigation.cs
@@ -1,0 +1,202 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Media;
+using SeeShells.UI.Pages;
+
+namespace SeeShells.UI.Node
+{
+    /// <summary>
+    /// Used to navigate trough nodes one by one
+    /// </summary>
+    public class NodeNavigation
+    {
+        private static NLog.Logger logger = NLog.LogManager.GetCurrentClassLogger();
+
+        ResourceDictionary resourceDictionary = new ResourceDictionary();
+        List<Object> nodes;
+        Object currNode;
+        int currIndex;
+
+        public NodeNavigation(List<Object> nodes)
+        {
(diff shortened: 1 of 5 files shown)


WPF/SeeShells/SeeShells/UI/Node/NodeNavigation.cs
```diff
@@ -0,0 +1,202 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Media;
+using SeeShells.UI.Pages;
+
+namespace SeeShells.UI.Node
+{
+    /// <summary>
+    /// Used to navigate trough nodes one by one
```
bridCquinn, line 13: Not a big deal, but misspelling of "through".

AlekStoyanov:
> There are a few spelling errors, and I feel like the color red clashes a bit, but nice job.

Spelling is fixed.
About the color, if you have any suggestions, I am more than happy to implement them.

WPF/SeeShells/SeeShells/UI/Node/NodeNavigation.cs
```diff
@@ -0,0 +1,198 @@
...
+            else
+            {
+                currIndex--;
+            }
+
+        }
+
+        /// <summary>
+        /// Navigates to the next node back in time.
+        /// </summary
```
klaki892, line 61: broken close tag

(found because github stopped color coding the code)

szczura: I noticed the different color & was so confused about it lol.

PR #332, opened 2020-04-10
Online parsing resolves actual user name before showing SID.

### LINKED ISSUE #192: Online hive should attempt to resolve actual user name before showing SID
_**Originally posted in #185 **_

Online username currently retrieve SID's because they are guaranteed and its non-trivial to resolve SID's to usernames without the SAM registry hive.

Because users of the program are not used to seeing the SID and instead the user name that owns the SID, things like the user filter(#164) would be better if the username was shown instead of the SID for each windows account.

 Offline parsing does this as of #185 and live parsing should use the same username locations as offline and fallback to SID's when not available.

Acceptance Criteria:
- When parsing a online registry, the system should attempt to retrieve the user's readable name. In the event it cant be determined fall back to the SID of the user account.

AlekStoyanov:
The `OnlineRegistryReader` tries to resolve the actual usernames of users first and uses SIDs if resolving failed.

image

Resolves #192

### DIFF
--- a/WPF/SeeShells/SeeShells/ShellParser/Registry/OnlineRegistryReader.cs
+++ b/WPF/SeeShells/SeeShells/ShellParser/Registry/OnlineRegistryReader.cs
@@ -62,10 +62,8 @@ namespace SeeShells.ShellParser.Registry
         {
             string retval = string.Empty;
 
-            //todo refactor this List into key-value pairs for lookup, we have to hardcode key-values otherwise.
             List<string> usernameLocations = Parser.GetUsernameLocations();
 
-            //todo we know of the Desktop value inside the "Shell Folders" location, so naively try this until a better way is found
             Dictionary<string, int> likelyUsernames = new Dictionary<string, int>();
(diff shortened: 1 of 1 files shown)


WPF/SeeShells/SeeShells/ShellParser/Registry/OnlineRegistryReader.cs
```diff
@@ -52,6 +58,52 @@ public List<RegistryKeyWrapper> GetRegistryKeys()
...
         }
 
+        private string FindOnlineUsername(RegistryKey userStore)
+        {
+            string retval = string.Empty;
+
+            //todo refactor this List into key-value pairs for lookup, we have to hardcode key-values otherwise.
+            List<string> usernameLocations = Parser.GetUsernameLocations();
+
+            //todo we know of the Desktop value inside the "Shell Folders" location, so naively try this until a better way is found
```
szczura, line 68: Might wanna get rid of to-dos since we're finishing the project.

PR #340, opened 2020-04-13
Website download

### LINKED ISSUE #21: Website Download program link
As an investigator or regular user, I want to click on a download link for SeeShells.

szczura:
Resolves #21 

Just need to remember to update the release or make a new release on GitHub. Code won't have to change since it points to the latest release :)


website/src/components/Home.vue
```diff
@@ -22,7 +22,12 @@
 
 <script>
 export default {
-  name: 'Home'
+        name: 'Home',
+        methods: {
+            onClick() {
+                window.open('https://github.com/RickLeinecker/SeeShells/releases/latest/download/SeeShells.exe');
```
klaki892, line 28: This doesn't feel okay to use.
[Mozilla's web documentation discourages it:](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#Usability_issues)
> Generally speaking, it is preferable to avoid resorting to window.open() for several reasons:
>  - Most modern desktop browsers offer tab-browsing, and tab-capable browser users overall prefer opening new tabs than opening new windows in a majority of webpage situations.
> - **_Most modern browsers offers a popup-blocking feature._**
>  ...

szczura: I added a title to try to follow best practices according to here: https://developer.mozilla.org/en-US/docs/Web/API/Window/open

Since the tab closes right away to start the download, some of the suggestions don't seem to apply here since they seem to be intended for pop-up windows that stay on the screen.

